PR #3, opened 2020-06-02
Multithreaded 'run' (and copy) code

### LINKED ISSUE #2: GSL + OpenMP support
It possibly makes sense to do these together.

From my reading, we're best off doing the gsl rng per-thread, because that means that the memory is allocated near the processor running it (creating one RNG per particle may end up being less efficient). My understanding of this is not great though (see [this post](https://lists.gnu.org/archive/html/help-gsl/2014-05/msg00014.html)) and if there was something more happily parallel that we can use I'd be up for bringing that in, even if that means using C++, which is going to be easy enough to add here now.

If we need to allocate the data per thread, then all the uses of `Calloc`/`Free` need to be replaced with `calloc`/`free` and we need to catch any failure to allocates then _after_ the parallel block throw an error with `Rf_error()` - no allocations using the R API can be done in the block. This causes some minor problems with the use of `user`, especially as in the odin model sometimes the size of data is determined here and some extra thought will be required, but we can cross that bridge when we come to it

If the constant back-and-forth between R and C, combined with however threads are spawned, gets too much we can tweak this a bit by providing the eventual likelihood/shuffle function as an R callback - just something to think about.

The official R docs on OpenMP support are here: https://cran.r-project.org/doc/manuals/r-devel/R-exts.html#OpenMP-support - this means that at some point (not necessarily in the first cut) we need to make compilation work with or without OpenMP.

Provided that #1 does not radically change, I wonder if changing the `model_update` definition from

```c
typedef void model_update(void *data, size_t step, const double *y,
                          double *ynext);
```

to something like

```c
typedef void model_update(void *data, size_t step, const double *y,
                          gsl_rng * rng, double *ynext);
```

would be what we need here. Adding it to the contents of `data` would be the other option but would require that `data` is distributed over threads in the same way?

johnlees:
Uses OpenMP and GSL to multithread the run stage (and the copy stage, which looked like it should also be fine).

I've put guards in which should stop compile errors when using e.g. `clang` to compile.

Questions: 
- Should `rng_init` and `rng_free` use `Calloc` and `Free` instead of `calloc` and `free`?
- Is the way of dealing with seed desirable, or do we want to expose this option to the user? Is the seed + thread_idx ok? (I thought this was probably better than asking for an array/list of seeds).
- How about the generator type? I guessed that this wouldn't really be of interest to the user, as long as it was internally consistent

Closes #2


src/interface.c
```diff
@@ -36,12 +36,21 @@ void particle_finalise(SEXP r_ptr) {
 SEXP r_particle_run(SEXP r_ptr, SEXP r_step_end) {
   particle *obj = (particle*) read_r_pointer(r_ptr, true);
   size_t step_end = (size_t) INTEGER(r_step_end)[0];
+  
+  // Note: running a single particle allocs a gsl rng each
```
johnlees, line 40: This section is awkward, and makes me think that perhaps we should get rid of the single particle interface? Or, as it's a useful test, discourage it?

richfitz:
Closing in favour of #4

PR #4, opened 2020-06-10
Basic C++ header-only implementation

### LINKED ISSUE #2: GSL + OpenMP support

richfitz:
A new shot at this, but using the xoshiro RNG and a header-only C++ implementation.  Still to come, not all necessarily in this PR

* [ ] There is a copy that I'd like to eliminate in `dust::Dust::run()` - we should be able to pass in the memory directly (#10)
* [ ] It should be possible to set the "starting state" (#6)
* [ ] It should be possible to control index_y (but how? and in what base?) (#8)
* [ ] An example that lives out of the package and is compiled as a package (~Gabor has a package to help with this I believe~ this package does not help with compiled code) (#9)
* [x] Add header guards
* [x] Add support for reset
* [x] Fetch full state
* [x] The binomial/poisson draws should be copied over
* [x] Remove explicit thread index from models
* [x] It should be possible to set the "starting step"
* [x] An example that lives out of the package and is compiled as a standalone dll
* [x] Remove bundled example from src/
* [x] An interface around the `.Call` uses
* [x] An interface around those functions (possibly R6)

I think that the remaining things can be done in future PRs as this is already immense; we can take care of the bottom 3 when we create a proper interface (you'll notice that there are no exported functions yet and so no docs!)

Closes #2


src/example_walk.cpp
```diff
@@ -0,0 +1,74 @@
...
+  obj->run(step_end);
+
+  const size_t n_state = obj->n_state();
+  const size_t n_particles = obj->n_particles();
+
+  std::vector<double> dat(n_state * n_particles);
+  obj->state(dat);
+
+  SEXP ret = PROTECT(allocMatrix(REALSXP, n_state, n_particles));
+  memcpy(REAL(ret), &dat[0], dat.size() * sizeof(double));
```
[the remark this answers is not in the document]
richfitz: Yeah this is the copy. It's not a huge deal for now though, as it's just a small wart and can be dealt with later (it's unlikely to be a big deal in terms of speed either).

I think that it can be done though by something like

```
template <typename it>
void state(const std::vector<size_t>& index_y, it end_state) const {
  for (size_t i = 0; i < index_y.size(); i++) {
    *(end_state + i) = _y[index_y[i]];
  }
}
```

which would work with both a pointer and an iterator...

You're 💯 on `.data()` - I knew there was a better way to do this but could not remember it

inst/include/dust/rng.hpp
```diff
@@ -0,0 +1,42 @@
...
+  std::vector<Xoshiro> _generators;
+};
+
+// Constructor builds vector of XOSHIRO providing non-overlapping
+// streams of random numbers
+inline RNG::RNG(const size_t n_threads, const uint64_t seed) {
+    Xoshiro rng(seed);
+    for (size_t i = 0; i < n_threads; i++) {
+        _generators.push_back(rng);
+        rng.jump();
```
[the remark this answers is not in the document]
richfitz: So, literally just replacing `rng.jump()` with `rng.long_jump()` - seems good to me, I can do that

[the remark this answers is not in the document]
johnlees: Yes - just means should you ever wish to call `rng.jump()` from a thread (it spawns more threads or needs more independent RNGs) it will work

src/example_walk.cpp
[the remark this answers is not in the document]
johnlees: Ah yes, was thinking of templating the wrong function - that looks good

inst/include/dust/dust.hpp
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  // This scheme means that if we have the same number of generators
+  // and threads then work for thread i occurs on generator i.
+  //
+  // If we have more generators than threads then each thread uses m =
+  // n_generators / n_threads (so if we had 8 generators and 2
+  // threads, each thread would use 4 generators).  The first thread
+  // will cycle alternate through the first m generators, the second
+  // thread will cycle through the second m generators, etc.
```
richfitz, line 91: To see how this might work in practice, see:

```r
n_generators <- 8
n_threads <- 2
m <- n_generators / n_threads
i <- 0:24

thread_idx <- 0
i %% m + thread_idx * m

thread_idx <- 1
i %% m + thread_idx * m

# ----

n_generators <- 8
n_threads <- 4
m <- n_generators / n_threads
i <- 0:24

thread_idx <- 0
i %% m + thread_idx * m

thread_idx <- 1
i %% m + thread_idx * m

thread_idx <- 2
i %% m + thread_idx * m

thread_idx <- 3
i %% m + thread_idx * m
```

inst/include/dust/rng.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+class RNG {
+public:
+  RNG(dust::Xoshiro generator) : _generator(generator) {}
+
+  double unif_rand() {
+    return _generator.unif_rand();
+  }
+
+  double runif(double min, double max) {
```
johnlees, line 19: Can we template `runif` and `rnorm` too, and set the default to double? Will be useful if/when they become floats

inst/include/dust/rng.hpp
```diff
@@ -0,0 +1,66 @@
...
+#include <random>
+#include <dust/xoshiro.hpp>
+#include <dust/distr/binomial.hpp>
+#include <dust/distr/poisson.hpp>
+
+namespace dust {
+
+class RNG {
+public:
+  RNG(dust::Xoshiro generator) : _generator(generator) {}
```
johnlees, line 13: Pass by ref?

inst/include/dust/dust.hpp
```diff
@@ -0,0 +1,140 @@
...
+template <typename T>
+class Dust {
+public:
+  typedef typename T::init_t init_t;
+  Dust(init_t data, size_t step, const std::vector<size_t> index_y,
+       const size_t n_particles, const size_t n_threads,
+       const size_t n_generators, const double seed) :
+    _index_y(index_y),
+    _n_threads(n_threads),
+    _rng(n_generators, seed) {
```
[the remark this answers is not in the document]
johnlees: I see this is in `util.hpp`

inst/template/Makevars
```diff
@@ -0,0 +1,3 @@
+# -*- makefile -*-
+PKG_CXXFLAGS=-I$(PATH_DUST_INCLUDE) -DHAVE_INLINE $(SHLIB_OPENMP_CXXFLAGS)
```
johnlees, line 2: Are CFLAGS needed too in case we ever use openmp there too? Or are we unlikely to ever do that

inst/template/interface.cpp
```diff
@@ -0,0 +1,90 @@
+#include <dust/dust.hpp>
+#include <dust/util.hpp>
+
+{{model}}
+
+extern "C" void {{name}}_finalise(SEXP ptr) {
```
johnlees, line 6: I think these `extern "C"` blocks may want `noexcept`

src/test_rng.h
```diff
@@ -0,0 +1,22 @@
+#include <R.h>
+#include <Rinternals.h>
+
+#ifdef __cplusplus
```
johnlees, line 4: Also noexcept here I think. Can be neater with a single define at the top:
```
#ifdef __cplusplus
#define EXTERN
#define EXCEPT
#else
#define EXTERN extern "C"
#define EXCEPT noexcept
#endif
```
then use `EXTERN` and `EXCEPT` around each function

tests/testthat/test-rng.R
```diff
@@ -0,0 +1,147 @@
...
+  nn <- rep(n, m)
+  pp <- rep(p, m)
+
+  ans1 <- .Call(Ctest_rng_binom, nn, pp, 1L, 1L)
+  ans2 <- .Call(Ctest_rng_binom, nn, pp, 1L, 1L)
+  expect_identical(ans1, ans2)
+
+  ## Should do this with much more statistical rigour, but this looks
+  ## pretty good.
+  expect_equal(mean(ans1), n * p, tolerance = 1e-3)
```
johnlees, line 61: Maybe comparing with `qbinom` would be better?

tests/testthat/test-rng.R
```diff
@@ -0,0 +1,147 @@
...
+  m <- 100000
+  lambda <- 5
+  ll <- rep(lambda, m)
+  ans1 <- .Call(Ctest_rng_pois, ll, 1L, 1L)
+  ans2 <- .Call(Ctest_rng_pois, ll, 1L, 1L)
+  ans3 <- .Call(Ctest_rng_pois, ll, 2L, 1L)
+  expect_identical(ans1, ans2)
+  expect_false(all(ans1 == ans3))
+
+  expect_equal(mean(ans1), lambda, 1e-2)
```
johnlees, line 130: and `qpois` here

inst/template/interface.cpp
```diff
@@ -0,0 +1,90 @@
...
+  size_t step = dust::util::as_size(r_step, "step");
+  size_t n_particles = dust::util::as_size(r_n_particles, "n_particles");
+  size_t n_threads = dust::util::as_size(r_n_threads, "n_threads");
+  size_t n_generators = dust::util::as_size(r_n_generators, "n_generators");
+  size_t seed = dust::util::as_size(r_seed, "seed");
+  dust::util::validate_n(n_generators, n_threads);
+
+  std::vector<size_t> index_y = {0};
+
+  dust::Dust<{{type}}> *d =
```
johnlees, line 29: Would wrapping this in a `shared_ptr` work, rather than `new` and `delete` in the finaliser? Or is that not easily compatible with the R interface?

inst/include/dust/distr/binomial.hpp
```diff
@@ -0,0 +1,134 @@
+#ifndef DUST_DISTR_BINOMIAL_HPP
```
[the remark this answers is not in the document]
richfitz: it's not portable, so we need old-style header guards for CRAN

inst/include/dust/dust.hpp
```diff
@@ -0,0 +1,140 @@
...
+  // and threads then work for thread i occurs on generator i.
+  //
+  // If we have more generators than threads then each thread uses m =
+  // n_generators / n_threads (so if we had 8 generators and 2
+  // threads, each thread would use 4 generators).  The first thread
+  // will cycle alternate through the first m generators, the second
+  // thread will cycle through the second m generators, etc.
+  void run(const size_t step_end) {
+    size_t m = _rng.size() / _n_threads;
+#pragma omp parallel num_threads(_n_threads)
```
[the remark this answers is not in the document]
richfitz: Can you explain more what you mean here (you can also use the "suggestion" feature in github to write out what you're thinking explicitly)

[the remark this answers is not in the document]
johnlees: `run()` would be written as
```suggestion
  void run(const size_t step_end) {
    size_t m = _rng.size() / _n_threads;
    #pragma omp parallel num_threads(_n_threads)
    {
      size_t thread_idx = 0;
#ifdef _OPENMP
      thread_idx = omp_get_thread_num();
#endif

      #pragma omp for schedule(static) ordered
      for (size_t i = 0; i < _particles.size(); ++i) {
        #pragma omp ordered
        _particles[i].run(step_end, _rng(i % m + thread_idx * m));
      }
    }
  }
```
Don't know if the indent of the `ifdef` should also match

inst/include/dust/dust.hpp
```diff
@@ -0,0 +1,140 @@
...
+
+  std::vector<double> _y;
+  std::vector<double> _y_swap;
+};
+
+template <typename T>
+class Dust {
+public:
+  typedef typename T::init_t init_t;
+  Dust(init_t data, size_t step, const std::vector<size_t> index_y,
```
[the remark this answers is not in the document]
richfitz: No advantage here to pass-by-reference because we'll be copying once

inst/include/dust/dust.hpp
```diff
@@ -0,0 +1,140 @@
...
+       const size_t n_generators, const double seed) :
+    _index_y(index_y),
+    _n_threads(n_threads),
+    _rng(n_generators, seed) {
+    for (size_t i = 0; i < n_particles; ++i) {
+      _particles.push_back(Particle<T>(data, step));
+    }
+  }
+
+  void reset(init_t data, size_t step) {
```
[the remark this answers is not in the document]
richfitz: Yup, they can be - added

inst/include/dust/dust.hpp
```diff
@@ -0,0 +1,140 @@
...
+    {
+      size_t thread_idx = 0;
+#ifdef _OPENMP
+      thread_idx = omp_get_thread_num();
+#endif
+
+#pragma omp for schedule(static) ordered
+      for (size_t i = 0; i < _particles.size(); ++i) {
+#pragma omp ordered
+        _particles[i].run(step_end, _rng(i % m + thread_idx * m));
```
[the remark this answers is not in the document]
richfitz: Moved to a private method

inst/include/dust/dust.hpp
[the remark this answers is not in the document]
richfitz: I've updated the pragma indenting

inst/template/interface.cpp
richfitz: Obsolete now that we're using Rcpp

tests/testthat/test-rng.R
richfitz: That leaves us without an equivalent test for the mean though - still worth it?

R/interface.R
```diff
@@ -0,0 +1,49 @@
+dust <- function(alloc, run, reset, state, step, classname = "dust") {
+  R6::R6Class(
+    classname,
+    cloneable = FALSE,
+
+    private = list(
+      cpp_alloc = alloc,
```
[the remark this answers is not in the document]
richfitz: Shuffle (as `reorder`) will be in the Rcpp version that superceeds this PR

inst/template/interface.cpp
richfitz: Obsolete given #13

src/test_rng.h
richfitz: Obsolete given #13

tests/testthat/test-rng.R
johnlees: Or we can do both? 
The QQ plot test is what I'm used to from statistical genetics, it might be more sensitive than the mean

PR #18, opened 2020-06-17
I11 interface

### LINKED ISSUE #11: Proper interface, with an exported function
Currently whole interface is internal-only

richfitz:
Adds a super basic interface and some docs. Things will change (for example the initial state might want accepting by both the constructor and reset and I think we want to reorder the args to the constructor so that step can have a default arg) but this should get us going for now.

There is a manual step (`make website`) that pushes docs to https://mrc-ide.github.io - see in particular https://mrc-ide.github.io/dust/reference/dust.html

Fixes #11


R/interface.R
```diff
@@ -45,6 +249,36 @@ dust <- function(alloc, run, reset, state, step, reorder, classname = "dust") {
...
+
+dust_workdir <- function(path) {
+  if (is.null(path)) {
+    path <- tempfile()
+  } else if (file.exists(path)) {
+    if (!is_directory(path)) {
+      stop(sprintf("Path '%s' already exists but is not a directory",
+                   path))
+    }
+    contents <- dir(path, recursive = TRUE, no.. = TRUE, include.dirs = FALSE)
```
johnlees, line 270: Is this the right setting for recursive? I just launched this with a close to top level directory and this caused it to hang

R/interface.R
```diff
@@ -45,6 +249,36 @@ dust <- function(alloc, run, reset, state, step, reorder, classname = "dust") {
 }
 
 
+dust_guess_type <- function(txt) {
```
johnlees, line 252: I like the simplicity of this, and thinking about it, feels like it wouldn't easily load the wrong thing.
If there is an easy to use C++ parser or similar for R, finding a class and checking that it has the necessary methods (size etc) might be nice here, rather than giving an error at the compiler stage

R/interface.R
richfitz: Good point - will think about a better way of doing this

R/interface.R
richfitz: This issue https://github.com/richfitz/RcppR6/issues/1 and this proof-of-concept https://github.com/richfitz/cppinfo show the rabbit-hole that this becomes

Basically compiling C++ (rather than C) is extremely difficult - libclang might be long-term possibility....

johnlees:
Another quick thought: add the URL of the docs to the README

PR #23, opened 2020-06-22
i17 add basic parallel test

### LINKED ISSUE #17: Test for openmp/parallel behaviour
Write a simple model that does something like walk, but into another state variable writes out the current thread if running in openmp, -1 otherwise. Then we can verify that openmp/parallelisation is working as expected.

This might be useful for #14 too

richfitz:
This PR adds a simple test of parallel behaviour. The test is disabled on systems where openmp is not supported. I added a small utility function that returns some core information about how openmp is configured that eventually we will pull out as an exported function, but for now it's just in utils.R (note the note about whether or not the _compiler_ supports openmp, which is different from if the _package_ supports it in the case it was installed from binary).

~(No API changes here so I've not bumped the version number)~

Fixes #17


inst/examples/parallel.cpp
```diff
@@ -0,0 +1,45 @@
...
+    double sd;
+  };
+  parallel(const init_t& data) : data_(data) {
+  }
+  size_t size() const {
+    return 2;
+  }
+  std::vector<double> initial(size_t step) {
+#ifdef _OPENMP
+    const bool has_openmp = true;
```
johnlees, line 18: I think this is `static`

richfitz:
(note this now builds off #30 so that should be merged first)

PR #28, opened 2020-06-25
i27: Compile time specification of numeric types

### LINKED ISSUE #27: Support float as an alternative to double
Not that bad - the classes become templated against a storage type - switching out the RNG things to support this would also be worthwhile

richfitz:
This PR removes explicit use of `double` and `int` from (almost) all functions in dust. This will break odin.dust but the fix there is relative small and can be done as https://github.com/mrc-ide/odin.dust/issues/4 (see https://github.com/mrc-ide/odin.dust/pull/5 for progress). It also breaks mcstate (see https://github.com/mrc-ide/mcstate/pull/18 for fix).

Things I'd like feedback on:

* The binomial.hpp function includes lots of use of `double` that I am not totally sure about templating away - I've not updated that blindly as I am not familiar with the algorithm but it looks like that should be ok?
* Both `double` and `int` are templated - if we only need a `double`/`float` distinction we could make this somewhat simpler
* There are no tests that the `float` interface actually works mostly because it's not obvious to me what a good test would look like here?
* It would be possible to make the `T::int_t` and `T::real_t` typedefs in the model class optional and default to `int` and `double` respectively. However, that will involve some fairly ugly template magic I think, plus I am leading to doing this explicitly everywhere to make it less easy to get wrong...

Fixes #27

### DIFF
--- a/inst/include/dust/distr/binomial.hpp
+++ b/inst/include/dust/distr/binomial.hpp
@@ -6,14 +6,14 @@
 namespace dust {
 namespace distr {
 
-template <typename real_t, typename int_t, typename rng_t>
-int_t binomial_inversion(int_t n, real_t prob, rng_t& generator) {
-  real_t geom_sum = 0;
-  int_t num_geom = 0;
+template <typename rng_t>
+double binomial_inversion(double n, double prob, rng_t& generator) {
+  double geom_sum = 0;
+  double num_geom = 0;
 
   while (true) {
-    real_t r = generator.unif_rand();
-    real_t geom = std::ceil(std::log(r) / std::log1p(-prob));
+    double r = generator.unif_rand();
+    double geom = std::ceil(std::log(r) / std::log1p(-prob));
     geom_sum += geom;
     if (geom_sum > n) {
       break;
@@ -23,17 +23,16 @@ int_t binomial_inversion(int_t n, real_t prob, rng_t& generator) {
   return num_geom;
 }
 
-template <typename real_t>
-real_t stirling_approx_tail(real_t k) {
-  static real_t kTailValues[] = {0.0810614667953272,  0.0413406959554092,
+inline double stirling_approx_tail(double k) {
+  static double kTailValues[] = {0.0810614667953272,  0.0413406959554092,
                                  0.0276779256849983,  0.02079067210376509,
                                  0.0166446911898211,  0.0138761288230707,
                                  0.0118967099458917,  0.0104112652619720,
                                  0.00925546218271273, 0.00833056343336287};
   if (k <= 9) {
-    return kTailValues[static_cast<short>(k)];
+    return kTailValues[static_cast<int>(k)];
   }
-  real_t kp1sq = (k + 1) * (k + 1);
+  double kp1sq = (k + 1) * (k + 1);
   return (1.0 / 12 - (1.0 / 360 - 1.0 / 1260 / kp1sq) / kp1sq) / (k + 1);
 }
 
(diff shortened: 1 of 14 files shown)


inst/include/dust/distr/binomial.hpp
```diff
@@ -23,22 +23,26 @@ double binomial_inversion(double n, double prob, RNG& generator) {
...
+template <typename real_t>
+real_t stirling_approx_tail(real_t k) {
+  static real_t kTailValues[] = {0.0810614667953272,  0.0413406959554092,
                                  0.0276779256849983,  0.02079067210376509,
                                  0.0166446911898211,  0.0138761288230707,
                                  0.0118967099458917,  0.0104112652619720,
                                  0.00925546218271273, 0.00833056343336287};
   if (k <= 9) {
-    return kTailValues[static_cast<int>(k)];
+    return kTailValues[static_cast<short>(k)];
```
johnlees, line 34: I hadn't come across `short` before. Usually I use `uint16_t` as I think it's very clear how much space is being used. This is probably moot though, I'm not even sure why this case is needed here

inst/include/dust/distr/binomial.hpp
```diff
@@ -23,22 +23,26 @@ double binomial_inversion(double n, double prob, RNG& generator) {
   return num_geom;
 }
 
-inline double stirling_approx_tail(double k) {
-  static double kTailValues[] = {0.0810614667953272,  0.0413406959554092,
+template <typename real_t>
+real_t stirling_approx_tail(real_t k) {
+  static real_t kTailValues[] = {0.0810614667953272,  0.0413406959554092,
```
johnlees, line 28: This function may want to stay as a double, but I'm not sure. In the tf code it isn't templated

inst/include/dust/distr/binomial.hpp
```diff
@@ -23,22 +23,26 @@ double binomial_inversion(double n, double prob, RNG& generator) {
...
   if (k <= 9) {
-    return kTailValues[static_cast<int>(k)];
+    return kTailValues[static_cast<short>(k)];
   }
-  double kp1sq = (k + 1) * (k + 1);
+  real_t kp1sq = (k + 1) * (k + 1);
   return (1.0 / 12 - (1.0 / 360 - 1.0 / 1260 / kp1sq) / kp1sq) / (k + 1);
 }
 
+// TODO: Can't template this until I understand why 'n' is coming in
```
johnlees, line 40: In the tf code this function uses `double` everywhere, and in the main loop `n` is typecast to double and `k` typecast after return. I think it's just to save writing potentially many type casts

inst/include/dust/distr/binomial.hpp
```diff
@@ -103,22 +107,22 @@ T rbinom(RNG& generator, int n, double p) {
...
-  if (n < 0 || p < 0 || p > 1) {
+    if (n < 0 || p < 0 || p > 1) {
     return NaN;
-  }
+    }
   */
 
-  double q = p;
+  real_t q = p;
   if (q > 0.5) {
```
johnlees, line 116: This should probably be p > 0.5 to be consistent with the conditional below

inst/include/dust/distr/binomial.hpp
richfitz: This reverts back to `int` now leaving the tf code largely untouched

inst/include/dust/distr/binomial.hpp
richfitz: yup, reverted now - I should have checked the tf code first!

PR #30, opened 2020-06-25
i24 progress towards getting arbitrary information out of the dust object

### LINKED ISSUE #24: Support for custom methods
Probably required for https://github.com/mrc-ide/odin.dust/issues/2 - allow custom methods to be added to the object

richfitz:
Smaller in scope than what is possible in odin, here we allow some information to be computed against the data on initialisation and reset.  This will be sufficient for odin.dust to feed back information about the way that the variables are ordered, and cope with them being resized due to parameter changes.

Must be merged after #28 and rebased/merged to get a clean diff once that's merged.

Requires work on odin.dust (https://github.com/mrc-ide/odin.dust/pull/8)

Fixes #24 (at least in the narrowest intent of that issue)

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: dust
 Title: Iterate Multiple Realisations of Stochastic Models
-Version: 0.0.2
+Version: 0.0.3
 Authors@R: c(person("Rich", "FitzJohn", role = c("aut", "cre"),
                     email = "rich.fitzjohn@gmail.com"),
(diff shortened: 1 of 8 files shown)


richfitz:
> Can we add `info` to the docs (this would be useful when writing this into mcstate), and possibly also to the sir.cpp example to show how it might relate to a typical odin model?

I've added this to the docs - however, with mcstate we'll be using `index` probably. Though now I am wondering if we should just pop that here too. There's a bit of an issue with the packaging interface getting additional methods across...

PR #31, opened 2020-06-30
i29: create packaged dust code

### LINKED ISSUE #9: Example use from a package

### LINKED ISSUE #29: Package interface
Generate code that can be dumped into src/ for a package (equivalent of `odin::odin_package`)

richfitz:
This PR is introduces the function `dust::dust_package` which will convert C++ files in `inst/dust` into dust models in `src/` and an R6 interface in `R/dust.R` if there are multiple models, then multiple C++ files are created.

Most of the work is ensuring that we're in a position to edit files at the given path - are they really dust files, will we overwrite anything etc. I've also added checks of the DESCRIPTION and NAMESPACE that should increase the chance that the created code actually works.

All fafftastic but gets the job done, I think.

There's additional work to come in `odin.dust` (https://github.com/mrc-ide/odin.dust/issues/9) that will use this interface (https://github.com/mrc-ide/odin.dust/pull/10)

Fixes #29 
Fixes #9


R/package.R
```diff
@@ -0,0 +1,157 @@
...
+  root <- package_validate(path)
+  path_dust <- file.path(root, "inst/dust")
+  path_src <- file.path(root, "src")
+  path_r <- file.path(root, "R")
+
+  ## 2. find target model files
+
+  ## TODO: support alternative location, perhaps not in package - that
+  ## will suit odin.
+  files <- dir(path_dust, pattern = "\\.cpp$")
```
johnlees, line 32: We may want to make this more general/user specifiable. `odin.dust` will want to output .cu (and possibly .cpp at the same time) when we're using CUDA

R/package.R
```diff
@@ -0,0 +1,157 @@
...
+  data <- lapply(file.path(path_dust, files), package_generate,
+                 template_cpp, template_r)
+
+  dir.create(path_src, FALSE, TRUE)
+  dir.create(path_r, FALSE, TRUE)
+  for (d in data) {
+    writeLines(d$cpp, file.path(path_src, d$filename))
+  }
+
+  code_r <- c("## Generated by dust - do not edit",
```
johnlees, line 52: It might be nice to add the dust version in here

R/package.R
richfitz: I'm in favour of crossing this bridge when we come to it. We'll probably have a bunch of things that come together to tie up a model and associated code

R/package.R
richfitz: done in i25-rng-interface

Issue #26, opened 2020-06-24
Add interface to long_jump function for RNG


johnlees:
This would be useful if dispatching jobs to different CPUs, or if trying to add another layer of parallelism. Then each process can safely call jump and spawn threads itself.

richfitz:
This would be good to sketch out the full interface for; I guess we're looking here for something on the dust object?

PR #32, opened 2020-06-30
i25 proper interface to the RNG

### LINKED ISSUE #25: Proper access to the RNG for testing
Having to go through private access in odin.dust tests. Needs some thoughts about interface though

richfitz:
This PR exposes the rng out as a first-class object, which will make testing easier I think.  Also will be useful for #26 if we start messing about with jumps (I've exposed them here)

Fixes #25


tests/testthat/test-rng.R
```diff
@@ -133,15 +143,79 @@ test_that("poisson numbers", {
...
+
+  m1 <- rbind(r1a, r1b, deparse.level = 0)
+  m2 <- matrix(r2, 2)
+  expect_equal(m1, m2)
+})
+
+
+test_that("long jump", {
+  seed <- 1
+  rng1 <- dust_rng$new(seed, 1L)
```
johnlees, line 210: I think the ideal test would be to check that the following are different:
```suggestion
  rng1 <- dust_rng$new(seed, 1L)$jump()
  rng2 <- dust_rng$new(seed, 1L)$long_jump()$jump()
```
As the use of `long_jump()` is so that you can call `jump()` on other processes, without needing to know how many jumps have been called

richfitz: ok cool - I added it as a 4th combination and test all comparisons of the 4

richfitz: (9028408)

PR #36, opened 2020-07-01
i8: allow setting index and state

### LINKED ISSUE #6: Run from given arbitrary state/time
Needed for projections

### LINKED ISSUE #8: Control for index_y
Because initialisation controls the size of the system we might want to set this either at initialisation/reset or after.  After seems most general and not that bad to implement.

So `reset` etc will set index_y to be all indices and then we have a "set index" method that will apply it

richfitz:
This PR allows setting the "index" (the filter applied to the state vector after run) and also the state.

There is some subtlety here. We do not know the *size* of the problem until after the model object has been created (i.e., `data_t` passed to the model constructor) and that is done in a block of code that should not throw.  So we set `data` and `step` here, but change `index` and `state` in a function that must be applied *after* this has happened. So either after a creation or a reset one can use `set_index` or `set_state`. The default index is now a much more sensible empty vector, default state is that implied by the model.

I have updated `index_y` to `index` everywhere as it did not feel that was adding much benefit. Later we may rename again to `index_run` in contrast with any other indices that we store (in the outside chance that checking the index is a bottleneck).

I've bumped the version number to 0.1.0 as this gets close to a MVP and should be stable enough to start work again on mcstate.

Fixes #8 
Fixes #6

### DIFF
--- a/R/interface.R
+++ b/R/interface.R
@@ -205,8 +205,6 @@ dust_interface <- R6::R6Class(
     ##' elements between 1 and the length of the state (this will be
     ##' validated, and an error thrown if an invalid index is given).
     set_index_y = function(index_y) {
-      .Call(private$cpp_set_index_y, private$ptr, index_y)
-      invisible()
     },
 
     ##' @description
@@ -217,8 +215,6 @@ dust_interface <- R6::R6Class(
     ##' @param state The state vector - must be a numeric vector with the
     ##' same length as the model's current state.
     set_state = function(state) {
-      .Call(private$cpp_set_state, private$ptr, state)
-      invisible()
     },
 
     ##' @description
--- a/R/interface.R
+++ b/R/interface.R
@@ -195,6 +195,33 @@ dust_interface <- R6::R6Class(
     },
 
     ##' @description
+    ##' Set the "index" vector that is used to return a subset of data
+    ##' after using `run()`. If this is not used then `run()` returns
+    ##' an empty (zero-row) matrix. This method must be called after any
+    ##' call to `reset()` as `reset()` may change the size of the state
+    ##' and that will invalidate the index.
+    ##'
+    ##' @param index_y The index vector - must be an integer vector with
+    ##' elements between 1 and the length of the state (this will be
+    ##' validated, and an error thrown if an invalid index is given).
+    set_index_y = function(index_y) {
+      .Call(private$cpp_set_index_y, private$ptr, index_y)
+      invisible()
+    },
+
+    ##' @description
+    ##' Set the "state" vector for all particles, overriding whatever your
+    ##' models `initial()` method provides. Currently all particles are set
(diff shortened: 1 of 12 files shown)


R/interface.R
```diff
@@ -261,11 +285,23 @@ dust_class <- function(alloc, run, reset, state, step, reorder,
         .Call(private$cpp_run, private$ptr, step_end)
       },
 
+      ## TODO: this probably needs changing as we can't set state and
```
johnlees, line 288: Can you add a test `is.null(private$data)` below? Or is NULL possible after data has been set?

R/interface.R
```diff
@@ -194,6 +194,29 @@ dust_interface <- R6::R6Class(
     run = function(step_end) {
     },
 
+    ##' @description
+    ##' Set the "index" vector that is used to return a subset of data
```
johnlees, line 198: A default value of 0 to n-1 might be more intuitive default behaviour than returning an empty matrix

richfitz: yeah, I think you're right there

R/interface.R
richfitz: Sorry, that was a stale comment!

inst/include/dust/dust.hpp
```diff
@@ -235,6 +229,22 @@ class Dust {
...
     return _rng(i % _rng.size());
   }
+
+  void initialise(const init_t data, const size_t step,
+                  const size_t n_particles) {
+    _particles.clear();
+    _particles.reserve(n_particles);
+    for (size_t i = 0; i < n_particles; ++i) {
+      _particles.push_back(Particle<T>(data, step));
+    }
```
johnlees, line 239: Just overwrite
```suggestion
    for (size_t i = 0; i < n_particles; ++i) {
      _particles[i] = Particle<T>(data, step);
    }
```

johnlees: Although this won't work if  `n_particles != _particles.size()` - is that the case sometimes?

inst/include/dust/dust.hpp
```diff
@@ -235,6 +229,22 @@ class Dust {
...
+
+    const size_t n = n_state_full();
+    _index.clear();
+    _index.reserve(n);
+    for (size_t i = 0; i < n; ++i) {
+      _index.push_back(i);
+    }
```
johnlees, line 246: Similar here, an overwrite in the loop would be simpler

inst/include/dust/dust.hpp
richfitz: Once created particle size is immutable

inst/include/dust/dust.hpp
richfitz: The index here *can* change size so we need to clear and push back so that the length is correct

PR #38, opened 2020-07-03
i33; less duplication in class definitions

### LINKED ISSUE #33: Avoid duplication of dust class details
There are 3 copies

* one used for documentation to exploit the R6 Roxygen support (`dust_interface`)
* one used for creating dust models in-session (`dust_class`)
* one used for creating dust models in a package (inst/template/dust.R.template)

At the least we should test that these all conform. It's possible that this can be achieved with inheritance but I think we'll get a speed hit there

richfitz:
This PR fixes some pretty egregious duplication and now there is but one source of truth for the R6 classes that dust uses. There *are* still two copies (down from 3) but one is generated from another and has a great big notice to this effect

Fixes #33


richfitz:
> dust_package is separate to usage through dust function, is that right? Plans to add a vignette for dust_package? I'm not very clear on how dust_package is supposed to be used.

I've added this to the main vignette

> Also looks like no test that the cache works?

Added a test of that too

Issue #7, opened 2020-06-11
Helper for run-and-collect


richfitz:
Needed for projections

PR #46, opened 2020-07-08
i45: allow variable initial time

### LINKED ISSUE #45: Allow initialisation with variable times and starting points
Like a burn-in, needed to support https://github.com/mrc-ide/mcstate/issues/25, and building off of #20/#23

Need a new function, perhaps call it "burn-in" that will take arguments:

```
burnin = function(step, state)
```

where `step_initial` is a vector of initial step sizes (same length as the number of particles, or length 1) and `state` is a matrix of initial states (same number of columns as the number of particles, or a vector). We initialise the system and run forward up to the largest value of `step`, each particle in turn, in parallel

richfitz:
This PR allows for a "burnin" period in particles, where we accept a variable initial time. This is needed for the sircovid work because we allow the precice seeding date to be drawn from a truncated poisson distribution. To do this we allow multiple a variable step but as part of the initialisation we run all particles up to the largest step given. No observation of the process can be added during this period so it's not possible to get a trace of the particles (probably something to consider when doing #7)

Fixes #45

### DIFF
--- a/vignettes/dust.Rmd
+++ b/vignettes/dust.Rmd
@@ -297,6 +297,55 @@ matplot(time, state, type = "l", lty = 1, col = "#00000022",
         xlab = "Time", ylab = "Number infected (I)")
 ```
 
+You can optionally set the initial step at the same time as the state. This is useful in two cases; if your model depends on time (e.g., you use the step in a calculation by transforming it into some measure of time) or if your particles start at different points. For example, suppose we start 10 individuals in the infected state but we do so with a number of possible starting points sampled from between step 0 and step 30:
+
+```{r initial_step}
+step0 <- sample(0:30, 20, replace = TRUE)
+model <- sir$new(list(), 0, 20)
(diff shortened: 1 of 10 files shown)


vignettes/dust.Rmd
```diff
@@ -297,6 +297,55 @@ matplot(time, state, type = "l", lty = 1, col = "#00000022",
...
+model$set_state(c(1000, 10, 0), step0)
+```
+
+After initialisation, our model is now at `r max(step0)`, the last step referenced in `step0`
+
+```{r}
+model$step()
+```
+
+and the initial states are a mix of states, despite having been seeded with the same 10 infected individuals, as different particles have run for the same time to reach this time point:
```
r-ash, line 314: different particles have run for different times to reach this time point?

richfitz: 🤦

richfitz: Fixed in ae39ed6

PR #59, opened 2020-07-17
i58: rng state

### LINKED ISSUE #58: Set and get state of the RNG

richfitz:
This PR adds a method `state` to the rng object and `rng_state` to the dust object, for retrieving the RNG state. Little useful can be done with these but they might be helpful for debugging.

Fixes #58

### DIFF
--- a/R/dust_class.R
+++ b/R/dust_class.R
@@ -123,7 +123,10 @@ dust_class <- R6::R6Class(
     },
 
     ##' @description
-    ##' Returns the state of the random number generator
+    ##' Returns the state of the random number generator. This returns a
+    ##' raw vector of length 32 * n_particles. It is primarily intended for
+    ##' debugging as one cannot (yet) initialise a dust object with this
+    ##' state.
     rng_state = function() {
     }
   ))
--- a/R/dust_class.R
+++ b/R/dust_class.R
@@ -120,5 +120,10 @@ dust_class <- R6::R6Class(
     ##' Only returns non-NULL if the model provides a `dust_info` template
     ##' specialisation.
     info = function() {
+    },
+
+    ##' @description
+    ##' Returns the state of the random number generator
+    rng_state = function() {
     }
   ))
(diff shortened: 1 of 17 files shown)


R/dust_class.R
```diff
@@ -120,5 +120,13 @@ dust_class <- R6::R6Class(
     ##' Only returns non-NULL if the model provides a `dust_info` template
     ##' specialisation.
     info = function() {
+    },
+
+    ##' @description
+    ##' Returns the state of the random number generator. This returns a
+    ##' raw vector of length 32 * n_particles. It is primarily intended for
```
johnlees, line 127: Specify that the vector is of positive/unsigned integers (and the other places this appears in docstrings)

inst/include/dust/xoshiro.hpp
```diff
@@ -42,6 +42,8 @@ class Xoshiro {
   void jump();
   void long_jump();
 
+  void get_state(std::vector<uint64_t>& state);
```
johnlees, line 45: Question rather than review comment: why do all the Dust functions return void, and we provide a vector of the correct size to insert results into? That feels quite C rather than C++ to me. Does it make it easier with the R interface/pointers? (perhaps I am just forgetting something we discussed when we first wrote it...)

R/dust_class.R
richfitz: From the point of view in R, it's only a raw vector (they're opaque) so it *represents* unsigned integers but it's really just data at this point

inst/include/dust/xoshiro.hpp
richfitz: if we were inserting into it we'd have to pass in the index as well - here're we're just appending. If I was doing it in C I'd probably do as you describe though and make this

```c
size_t get_state(uint64_t *state, size_t index) {
  for (size_t i = 0; i < XOSHIRO_WIDTH; ++i, ++index) {
    state[index] = _state[i];
  }
  return index;
}
```

PR #67, opened 2020-07-23
i60: Refactor RNG state

### LINKED ISSUE #60: Replace rng object with state that is passed around
In order to match the gpu version

```
git diff --no-index dust/inst/include/dust/xoshiro.hpp dust-gpu-prototype/src/xoshiro.hpp 
```

richfitz:
This PR is a bit painful - it refactors the RNG state so that we pass an opaque state object into free functions. This should create a fairly similar interface for the CPU and GPU versions.

Note that I removed `int_t` templating as it was not needed and complicated some of the templates.

Of note:

* inst/include/dust/distr/normal.hpp throws away the second normal (for now)
* inst/include/dust/xoshiro.hpp includes the definition of rng_state_t which can eventually hold the 2nd normal
* The rng_state_t type is templated on our "real" type which prevents the user needing to specify that for the functions in dust::distr

Fixes #60


inst/examples/parallel.cpp
```diff
@@ -25,9 +25,10 @@ class parallel {
     return ret;
   }
   void update(size_t step, const std::vector<double>& state,
-              dust::RNG<double, int>& rng, std::vector<double>& state_next) {
+              dust::rng_state_t<real_t>& rng_state,
```
johnlees, line 28: `step`, `state`, `state_next`, `rng_state` felt like a more natural order to me

inst/examples/sir.cpp
```diff
@@ -23,17 +23,18 @@ class sir {
...
     double S = state[0];
     double I = state[1];
     double R = state[2];
     double N = S + I + R;
 
     double p_SI = 1 - std::exp(-(data_.beta) * I / N);
     double p_IR = 1 - std::exp(-(data_.gamma));
-    double n_IR = rng.rbinom(round(I), p_IR * data_.dt);
-    double n_SI = rng.rbinom(round(S), p_SI * data_.dt);
+    double n_IR = dust::distr::rbinom(rng_state, I, p_IR * data_.dt);
```
johnlees, line 36: Why did the `round` go?

tests/testthat/test-rng.R
```diff
@@ -64,6 +64,16 @@ test_that("run uniform random numbers with odd bounds", {
 })
 
 
+test_that("distribution of uniform numbers", {
```
johnlees, line 67: Do we have an `rnorm` test too?

inst/examples/sir.cpp
richfitz: Was in the generated odin code (where it's generally needed) but it's not actually needed here so I dropped it in pursuit of a simpler example

tests/testthat/test-rng.R
richfitz: there is one:

```r
test_that("norm_rand agrees with rnorm", {
  n <- 100000
  ans <- dust_rng$new(2, 1)$norm_rand(n)
  expect_equal(mean(ans), 0, tolerance = 1e-2)
  expect_equal(var(ans), 1, tolerance = 1e-2)
  expect_gt(ks.test(ans, "pnorm")$p.value, 0.1)
})
```

but I can add one that changes parameters too

inst/examples/parallel.cpp
richfitz: I always thought that it was "best practice" (whatever that means) to put your writable variables last. We don't *directly* write to rng_state (even though it's modified). I don't feel strongly though - what's your reasoning? Just grouping the state variables?

johnlees: Ah right, I guess I had never come across that rule for C/C++ before. I usually put them in what I consider to be a logical order, so yeah grouping state, grouping rng

PR #69, opened 2020-07-23
Generate GPU code as well as CPU code

### LINKED ISSUE #68: Add GPU-compatible code
See https://github.com/mrc-ide/sircovid-gpu-hack/pull/4/files for prototype

richfitz:
This PR starts our GPU journey. There are unresolved issues here, which should be moved into new issues:

* The update() signature is slightly different and needs fixing (probably in the CPU version?)
* The xoshiro.hpp header needs refactoring to simplify it and bring it in line with the cpu version
* we should look at normal.hpp as it's again pretty different
* ~The global `typedef float real_t` needs to be removed for us to work with double precision (you get confusing compilation warnings otherwise)~
* Collect up as much code as possible into shared headers (see [comment below](#issuecomment-663391258) for details here)

Testing remains hard to do, but we'll hammer a bunch of this out over the next few days.

~At the least we should check that we can compile (https://github.com/mratsim/Arraymancer/issues/30, https://github.com/jeremad/cuda-travis)~ We now check that the basic model can compile; this was a faff but I think very useful to have. See the build status of 7242004 below, which includes a deliberate break to the code used on the GPU - this fails to compile on travis, breaking the build. The code in adjacent commits 38e0eec and f7ff30e passes on travis.

Fixes #68


richfitz:
Summary of remaining differences and opportunities for harmonisation:

`dust.cpp`/ `gpu/dust.cu` : almost identical, could be generated from each other

```
git diff --no-index inst/template/dust.cpp inst/template/gpu/dust.cu
```

`gpu/dust.hpp` : has no counterpoint, but not could be generated

`dust.hpp`: lots of changes:

```
git diff --no-index inst/include/dust/dust.hpp inst/include/dust/gpu/dust.hpp
```

* destructors and memory management, offload to GPU in run. It's possible we could handle much of this via inheritance though?

* rng.hpp:

```
git diff --no-index inst/include/dust/rng.hpp inst/include/dust/gpu/rng.hpp
```

* Quite a few differences, only some of which are important. Refactoring xoshiro.hpp might help here

* xoshiro.hpp

```
git diff --no-index inst/include/dust/xoshiro.hpp inst/include/dust/gpu/xoshiro.hpp
```

This can be refactored to be basically identical to the cpu version, but made GPU aware

* distr/binomial.hpp and friends

```
git diff --no-index inst/include/dust/distr/binomial.hpp inst/include/dust/gpu/distr/binomial.hpp
git diff --no-index inst/include/dust/distr/poisson.hpp inst/include/dust/gpu/distr/poisson.hpp
git diff --no-index inst/include/dust/distr/uniform.hpp inst/include/dust/gpu/distr/uniform.hpp
```

* std::log vs log etc
* some gpu annotations (`__syncwarp` and `__device__`)
* elimination of some early exits
* our main rng entrypoint changes from `dust::unif_rand<real_t>` to `device_unif_rand` (an issue for all files)

```
git diff --no-index inst/include/dust/distr/normal.hpp inst/include/dust/gpu/distr/normal.hpp
```

* distr/normal.hpp - in addition we've got divergence here due to the partially completed buffer

inst/include/dust/gpu/cuda.cuh
```diff
@@ -0,0 +1,28 @@
+#ifndef DUST_CUDA_CUH
+#define DUST_CUDA_CUH
+#include <stdio.h>
+
```
johnlees, line 4: Add a note on how this file is used: include at the top of each `.cu` file (not in the `.hpp`/`.cuh` files in which `CUDA_CALL` is actually used, as this gives a compiler error)

inst/include/dust/gpu/distr/binomial.hpp
```diff
@@ -0,0 +1,150 @@
...
+  while (true) {
+    double r = device_unif_rand(rng_state);
+    double geom = ceil(log(r) / log1p(-prob));
+    geom_sum += geom;
+    if (geom_sum > n) {
+      break;
+    }
+    ++num_geom;
+  }
+  __syncwarp(); // CHECK: ok to call this here, when some threads in warp
```
johnlees, line 25: We can remove this comment - empirically these `__syncwarp();` didn't deadlock

inst/include/dust/gpu/rng.hpp
```diff
@@ -0,0 +1,144 @@
...
+struct RNGptr {
+  uint64_t* state_ptr;
+  unsigned int state_stride;
+  unsigned int particle_stride;
+};
+
+// Read state from global memory
+template <typename T>
+__device__
+rng_state_t<T> loadRNG(RNGptr& rng_state, int p_idx) {
```
johnlees, line 21: I would suggest changing the name of the `RNGptr` argument in this function to `global_memory_rng_state` or `d_rng_state` to make `rng_state` less overloaded, and be clearer about where in memory it resides

inst/include/dust/gpu/rng.hpp
```diff
@@ -0,0 +1,144 @@
...
+  rng_state_t<T> state;
+  for (int i = 0; i < XOSHIRO_WIDTH; i++) {
+    state.s[i] = rng_state.state_ptr[p_idx * rng_state.particle_stride +
+                                      i * rng_state.state_stride];
```
johnlees, line 25:
```suggestion
                                     i * rng_state.state_stride];
```

inst/include/dust/gpu/rng.hpp
```diff
@@ -0,0 +1,144 @@
...
+  }
+  return state;
+}
+
+// Write state into global memory
+template <typename T>
+__device__
+void putRNG(rng_state_t<T>& rng, RNGptr& rng_state, int p_idx) {
```
johnlees, line 33: Ditto with `rng_state` variable here

inst/include/dust/gpu/rng.hpp
```diff
@@ -0,0 +1,144 @@
...
+  for (int i = 0; i < XOSHIRO_WIDTH; i++) {
+    rng_state.state_ptr[p_idx * rng_state.particle_stride +
+                        i * rng_state.state_stride] = rng.s[i];
+  }
+}
+
+template <typename real_t>
+class pRNG { // # nocov
```
johnlees, line 41: Still want `nocov`?

inst/include/dust/gpu/rng.hpp
```diff
@@ -0,0 +1,144 @@
...
+    for (int i = 0; i < n; i++) {
+      _rngs.push_back(rng);
+      rng.jump();
+    }
+
+    CUDA_CALL(cudaMalloc((void** )&_d_rng_state.state_ptr,
+                          n * XOSHIRO_WIDTH * sizeof(uint64_t)));
+    // Set to be interleaved
+    _d_rng_state.state_stride = n;
+    _d_rng_state.particle_stride = 1;
```
johnlees, line 54: Add `assert(_d_rng_state.state_stride * _d_rng_state.particle_stride == n)` or something better if you don't like asserts

inst/include/dust/gpu/rng.hpp
```diff
@@ -0,0 +1,144 @@
...
+    return state;
+  }
+
+private:
+  // delete move and copy to avoid accidentally using them
+  pRNG ( const pRNG & ) = delete;
+  pRNG ( pRNG && ) = delete;
+
+  void put_state_device() {
+    std::vector<uint64_t> interleaved_state(size() * XOSHIRO_WIDTH);
```
johnlees, line 103: Change variable name to `flattened_state` (as it is not necessarily interleaved)

inst/examples/sirs.cpp
```diff
@@ -0,0 +1,88 @@
...
+  void update(size_t step, const std::vector<real_t>& state,
+              dust::rng_state_t<real_t>& rng_state,
+              std::vector<real_t>& state_next) {
+#endif
+    real_t S = state[0];
+    real_t I = state[1];
+    real_t R = state[2];
+    real_t N = S + I + R;
+
+    real_t p_SI = 1 - exp(- internal.beta * I / (real_t) N);
```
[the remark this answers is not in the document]
richfitz: Seems to work, yeah :)

inst/include/dust/dust.hpp
```diff
@@ -133,30 +138,30 @@ class Dust {
   }
 
   void run(const size_t step_end) {
-    #pragma omp parallel for schedule(static) num_threads(_n_threads)
+#pragma omp parallel for schedule(static) num_threads(_n_threads)
```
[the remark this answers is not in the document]
richfitz: I was matching the indent from the code I imported from you - this was a change to minimise the cpu/gpu diff

inst/include/dust/gpu/cuda.cuh
richfitz: Sorry, I don't think I understand this. Is how it's being used ok? If so do I just add:

```
// Include this file at the top of each .cu file (not in the .hpp/.cuh files in which CUDA_CALL is actually used, as this gives a compiler error)
```

in this file?

inst/include/dust/gpu/rng.hpp
richfitz: This was needed to avoid a false negative in the covr output (in the CPU version - this file is invisible to coverage checking as it's not compiled into the package binary). I've no idea why it was needed though

inst/include/dust/gpu/rng.hpp
richfitz: done in b8ffc38

inst/include/dust/gpu/distr/binomial.hpp
richfitz: done in 529ecde

inst/include/dust/dust.hpp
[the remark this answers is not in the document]
richfitz: (My editor is pretty enthusiastic about indenting this all the way the first column)

inst/include/dust/gpu/rng.hpp
richfitz: Added in 57477d4

inst/include/dust/gpu/rng.hpp
richfitz: 257aa11

inst/include/dust/gpu/rng.hpp
richfitz: 257aa11

inst/include/dust/dust.hpp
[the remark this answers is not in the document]
richfitz: re-indented (in both files) in d4253e4

inst/include/dust/gpu/cuda.cuh
richfitz: Hmm, I have included this in a .hpp file (in order to get the .cu file and the .cpp file as similar as possible). Is this erroring for you?

johnlees: Previously when I did this I got an error because I was ending up with two compile units (which is why I ended up adding relocatable device code, and all the mess that entailed). Perhaps I had included it in both the header and the .cu, not sure. If it's working then no worries

PR #70, opened 2020-07-24
Prototype to use cub's device select

johnlees:
Note to work this will need:
- ~Compilation with C++14~
- cub installed (comes with cuda >=11.0, but on raplab we will need to install it ourselves)


johnlees:
This compiles now

johnlees:
Requires CUDA 11 or cub install separately (https://github.com/NVlabs/cub/releases) - we will need to update the CI to get this to pass

PR #80, opened 2020-07-31
Add code for rbinom btrd algorithm

johnlees:
~Untested as yet!~

TODO: Test that all exits from the function are possible (otherwise likely no speedup over btrs, and a likely mistake in the implementation of these sections)


johnlees:
Interesting that the stirling values are different, here is how I generated them:
```
#include <cstdio>

double stirling_approx_tail(double k) {
   double kp1sq = (k + 1) * (k + 1);
  return (1.0 / 12 - (1.0 / 360 - 1.0 / 1260 / kp1sq) / kp1sq) / (k + 1);
}

int main(int argc, char* argv[]) {
   for (int i = 0; i <= 128; i++) {
      printf("%.17g,\n", stirling_approx_tail(static_cast<double>(i)));
   }

   return 0;
}
```

johnlees:
> Interesting that the stirling values are different, here is how I generated them:
> 
> ```
> #include <cstdio>
> 
> double stirling_approx_tail(double k) {
>    double kp1sq = (k + 1) * (k + 1);
>   return (1.0 / 12 - (1.0 / 360 - 1.0 / 1260 / kp1sq) / kp1sq) / (k + 1);
> }
> 
> int main(int argc, char* argv[]) {
>    for (int i = 0; i <= 128; i++) {
>       printf("%.17g,\n", stirling_approx_tail(static_cast<double>(i)));
>    }
> 
>    return 0;
> }
> ```

It might be because the first values were evaluated using a more accurate formulae/numerical integral

johnlees:
Here is the correct code:
```
#include <cstdio>
#include <cmath>

double stirling_approx_tail(int k) {
   return(std::lgamma(k + 1) - (std::log(std::sqrt(2 * M_PI)) + (k + 0.5) * std::log(k + 1) - (k + 1)));
   //int kp1sq = (k + 1) * (k + 1);
    //return (1.0 / 12 - (1.0 / 360 - 1.0 / 1260 / kp1sq) / kp1sq) / (k + 1);
}

int main(int argc, char* argv[]) {
  // double f0 = 0.0810614667953272;
   for (int i = 0; i <= 128; i++) {
      printf("%.17g,\n", stirling_approx_tail(i));
   }

   return 0;
}
```

johnlees:
QQ plot at larger np (n = 5000, p = 0.1)
MicrosoftTeams-image

johnlees:
btrs is faster (perhaps as our runif draws are faster now)

PR #86, opened 2020-08-12
i84: multi parameter simulation

### LINKED ISSUE #84: Multiparameter support
We'll need this to do projections efficiently, but it is also needed for working with SMC^2 etc.

Currently we allocate our dust object with

```c++
cpp11::list dust_alloc(cpp11::list r_data, int step,
                       int n_particles, int n_threads,
                       int seed);
```

which creates a new dust data object like:

```c++
typename T::init_t data = dust_data<T>(r_data);
```

then takes advantage of being able to copy that through the particles when we do

```c++
new Dust<T>(data, step, n_particles, n_threads, seed);
```

So if we passed in a list of data objects (realistically a second `dust_alloc` object), perhaps with a list and an index indicating how it would be reused. Then we do:

```c++
std::vector<typename T::init_t> data;
for (size_t i = 0; i < r_data.size(); ++i) {
  data.push_back(dust_data<T>(r_data[i]));
}
```

(or modern C++ equivalent). Then write a constructor that takes this vector-of-data, which mostly impacts part of the `initialise` method, with

```c++
for (size_t i = 0; i < n_particles; ++i) {
  _particles.push_back(Particle<T>(data, step));
}
```

becoming

```c++
for (size_t i = 0; i < n_particles; ++i) {
  _particles.push_back(Particle<T>(data[i], step));
}
```

richfitz:
This PR adds support for multiparameter simulation. This is needed to support mcstate's projection, needed to complete the sircovid2 port.

Coming out of the end of an MCMC chain we will have a matrix of final states (and a sampled bit of trajectory that got us there), and a matrix of parameters that we can transform. So we want something that can, in parallel, do a simple projection forwards in time for all of these.

If we had a generic "DustMulti" object this would be better, and we'll implement that for SMC2 and then possibly backport this to use it. For now though, this should do the trick.

There's some faff in here: we need not start from a model object itself, and usually it will be better not to as nothing changes in the model. However, we need to find the correct environment to get the generated code and associated dll.

**NOTE for reading the diff**: The previous `dust::dust_simulate` utility (and file) have been renamed to `dust::dust_iterate`

Fixes #84


inst/include/dust/interface.hpp
```diff
@@ -40,6 +44,44 @@ cpp11::list dust_alloc(cpp11::list r_data, int step,
   return cpp11::writable::list({ptr, info});
 }
 
+// think about rng seeding here carefully; should accept either a raw
+// vector or an integer I think.
+template <typename T>
+cpp11::writable::doubles dust_simulate(cpp11::sexp r_steps,
```
johnlees, line 50: I'm not familiar with the cpp11 standard, but are copies the right thing here, or would a reference be better? (are they pointers already?)

inst/include/dust/interface.hpp
```diff
@@ -40,6 +44,44 @@ cpp11::list dust_alloc(cpp11::list r_data, int step,
   return cpp11::writable::list({ptr, info});
 }
 
+// think about rng seeding here carefully; should accept either a raw
+// vector or an integer I think.
```
johnlees, line 48: What does this mean? Planning to send an rng_state through here rather than a seed, eventually?

inst/include/dust/interface.hpp
```diff
@@ -40,6 +44,44 @@ cpp11::list dust_alloc(cpp11::list r_data, int step,
...
+template <typename T>
+cpp11::writable::doubles dust_simulate(cpp11::sexp r_steps,
+                                       cpp11::list r_data,
+                                       cpp11::doubles_matrix r_state,
+                                       cpp11::sexp r_index,
+                                       const size_t n_threads,
+                                       const size_t seed) {
+  typedef typename T::real_t real_t;
+  std::vector<size_t> steps = validate_size(r_steps, "steps");
+  // TODO: broken for floats
```
johnlees, line 58: because cpp11 is `doubles_matrix` (and R doesn't have floats)? Could we cast everything at the right point?
One thing I've never worked out in C++ is using conditionals based on a template type (maybe you aren't supposed to be able to do that for a reason)

inst/include/dust/interface.hpp
```diff
@@ -229,8 +266,11 @@ inline std::vector<size_t> validate_size(cpp11::sexp r_x, const char * name) {
   return x;
 }
 
-inline std::vector<size_t> r_index_to_index(cpp11::sexp r_index,
-                                            size_t nmax) {
+inline std::vector<size_t> r_index_to_index(cpp11::sexp r_index, size_t nmax) {
```
johnlees, line 269: This and following function have confusing names, and also I cannot immediately work out their purpose - can we add a comment above both perhaps with their intended use?

inst/include/dust/interface.hpp
```diff
@@ -245,6 +285,15 @@ inline std::vector<size_t> r_index_to_index(cpp11::sexp r_index,
   return index;
 }
 
+inline std::vector<size_t> r_index_to_index_default(size_t n) {
+  std::vector<size_t> index;
+  index.reserve(n);
+  for (size_t i = 0; i < n; ++i) {
+    index.push_back(i);
+  }
```
johnlees, line 293: versus:
```
std::vector<size_t> index(n);
for (size_t i = 0; i < n; ++i) {
  index[i] = i;
}
```
which is preferred? Your way doesn't zero initialise/default construct the vector, but I often use mine as it's easier to use with openmp

johnlees: Also my preferred way of doing this:
```
#include <numeric>
std::vector<size_t> index(n);
std::iota(index.begin(), index.end(), 0);
```

johnlees: Lots of options for this trivial part of the PR it turns out

vignettes/dust.Rmd
```diff
@@ -210,12 +210,12 @@ model$state(c(S = 1L, R = 3L))
 
 Again, this copies names from the index if they are present.
 
-In order to run the simulation beginning-to-end, we use `dust::dust_simulate()`; this function is purely a convenience function and the intention is that you would write your own (see [`mcstate`](https://mrc-ide.github.io/mcstate) for example).
+In order to run the simulation beginning-to-end, we use `dust::dust_iterate()`; this function is purely a convenience function and the intention is that you would write your own (see [`mcstate`](https://mrc-ide.github.io/mcstate) for example).
```
johnlees, line 213: Do we need to add `dust_simulate()` to the vignette? Or is that premature

R/dust_class.R
```diff
@@ -122,6 +139,11 @@ dust_class <- R6::R6Class(
     info = function() {
     },
 
+    ##' @description
+    ##' Returns the `data` object that your model was constructed with.
+    data = function() {
```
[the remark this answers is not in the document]
richfitz: yeah, I went back and forth on this and for now, if you hold a dust object in R space it corresponds to one data entry. We can't do a mix without breaking type-safety all over the show so will do this properly later

inst/include/dust/dust.hpp
```diff
@@ -4,6 +4,7 @@
 #include <dust/rng.hpp>
 
 #include <algorithm>
+#include <sstream>
```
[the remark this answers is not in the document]
richfitz: I would have much preferred sprintf-style interpolation - I really am not a fan of C++ string streams and output streams!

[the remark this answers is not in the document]
johnlees: I think you did it the 'right' way for C++, though sprintf would have been tempting

R/simulate.R
```diff
@@ -1,57 +1,94 @@
...
+##' state.  For our immediate use case this is for simulating at the
+##' end of an MCMC where want to generate a posterior distribution of
+##' trajectories.
 ##'
-##' @param model A model, compiled with `dust::dust()` and initialised
+##' @section Random number generation:
 ##'
-##' @param steps A vector of steps - the first step must be same as
-##'   the model's step (i.e., `$step()`)
+##' This implementation leaves a number of issues that we will resolve
```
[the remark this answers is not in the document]
richfitz: The issue is the roundtrip. So we have our particle filter which has n particles each with an RNG state and we want to simulate m trajectories.

So going from filter -> predict we should grab some representation of the RNG state (either the first particle's state or the whole vector and set it into the object doing the simulation. If n < m or if we only take the first particle we'll do a bunch of jumps on the rest - that's fine, but needs work in dust.

Then once we're done with the simulation we need to pull this state back.

I guess what you're suggesting is to do a long_jump on the dust object after copying its state so that it's "moved along" and copy some state over into the bit doing the prediction?

In any case, I think that there are sufficient judgement calls here we should work through this on a separate PR!

inst/include/dust/interface.hpp
richfitz: yes - turns out my ideas on this are still mostly in my head so dumped out some here: #87

inst/include/dust/interface.hpp
richfitz: they're pointers already

inst/include/dust/interface.hpp
richfitz: ha, it's no longer broken for floats as I fixed that

inst/include/dust/interface.hpp
richfitz: sure, added that in 5b3cea7

inst/include/dust/interface.hpp
richfitz: I think they're all fine - with differences that probably become apparent if the code is in a performance-critical section. This will likely never get called the way that mcstate is shaping up! Do you feel strongly about the differences? In my mind the way it is clearest (least book-keeping, though the reserve micro-optiomisation could be removed), but I don't feel strongly at all

vignettes/dust.Rmd
richfitz: Definitely it should be added but perhaps we should go through and refresh the whole thing once I can drag this through mcstate and into sircovid2?

R/simulate.R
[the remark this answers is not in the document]
richfitz: As I've added it now, see #87

richfitz:
hmm, something along here is causing segfaults, starting at 52155c1b95f1dca32d95b16ded00f5866b09ace6 which suggests that I might have done something wrong with openmp?

inst/include/dust/dust.hpp
```diff
@@ -233,4 +234,47 @@ class Dust {
...
+  dust::pRNG<real_t> rng(n_particles, seed);
+  std::vector<real_t> ret(n_particles * n_state_return * steps.size());
+  size_t n_steps = steps.size();
+
+#pragma omp parallel for schedule(static) num_threads(n_threads)
+  for (size_t i = 0; i < particles.size(); ++i) {
+    particles[i].set_state(state.begin() + n_state_full * i);
+    for (size_t t = 0; t < n_steps; ++t) {
+      particles[i].run(steps[t], rng.state(i));
+      size_t offset = t * n_state_return * n_particles + i * n_state_return;
```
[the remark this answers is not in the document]
richfitz: clearer but more segfaulty 😁 - I had a bunch of segfaults when getting the bookeeping correct here, reverting this fixes it so I think I'll leave as-is

inst/include/dust/interface.hpp
johnlees: Yeah keep it as is, just wanted to write them out to think if it made much of a difference as I'd done this a lot recently (the zero init is the main difference with a big object)

inst/include/dust/dust.hpp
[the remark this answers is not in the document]
johnlees: I didn't think about the fact that this was inside an openmp block. That being said, I'm not sure exactly why it is segfaulting, but let's leave that mystery unsolved

PR #88, opened 2020-08-13
i87: Improve seed handling behaviour

### LINKED ISSUE #85: Tie our seed to R's
Default seed behaviour could move to `sample.int(.Machine$integer.max, 1)`

### LINKED ISSUE #87: Better RNG state handling
Global RNG state is nice because it updates everywhere. We don't have that so we should think about it more.

Something like 

```
dust_rng_state()
```

as a free function is needed to check an initial state. Options should be a raw vector (the results of getting the rng state from a dust object), an integer (set seed as currently) or `NULL` in which case follow the idea in #85

richfitz:
This PR improves seed handling behaviour. Previously we had a default of `seed = 1L` everywhere and the user had to manually provide a good seed to get different results. Not ideal in real use. But we already have a decent source of random numbers, particularly where we just want one: R's RNG.

Here, the `seed` argument to `dust` and `dust_simulate` allows 3 types:

* `NULL`: seed using R's RNG by sampling an integer on `0..INT_MAX`
* single integer: as before, seed with a user-provided integer
* raw vector: interpret as a `std::vector<uint64_t>` and seed as many of the RNGs as possible with this. Then use `jump()` for any leftovers.

In order to better support parent/child relationships between rng streams, reading the rng state can get the state to long jump. This means we pass the current state and update and the two streams are uncoupled. We have to use a long jump here as we already use a jump for the rng streams within a set.

Fixes #87 
Fixes #85


richfitz:
OK, in the end I've got the gpu build compiling at least 63e0469...3ec5d94, of which 17f3c78 is an objective win, so it can live....for now at least 😈

inst/include/dust/rng_interface.hpp
```diff
@@ -0,0 +1,38 @@
...
+#include <R_ext/Random.h>
+
+template <typename T>
+std::vector<uint64_t> as_rng_seed(cpp11::sexp r_seed) {
+  auto seed_type = TYPEOF(r_seed);
+  std::vector<uint64_t> seed;
+  if (seed_type == INTSXP || seed_type == REALSXP) {
+    int seed_int = cpp11::as_cpp<int>(r_seed);
+    if (seed_int <= 0) {
+      cpp11::stop("'seed' must be non-negative");
```
johnlees, line 13: A negative seed might be used, I've certainly seen -1 used before. Can we not (quite safely for this purpose) cast the int as an unsigned int, and also thereby retain the full possible range [0-2^64) for the seed?

tests/testthat/test-interface.R
```diff
@@ -145,8 +145,25 @@ test_that("get rng state", {
...
   obj <- res$new(list(sd = 1), 0L, np, seed = seed)
   expect_identical(
-    obj$rng_state(),
+    obj$rng_state(FALSE),
     dust_rng$new(seed, np)$state())
+  ## State does not update if called with advance = FALSE
+  expect_identical(
+    obj$rng_state(FALSE),
+    dust_rng$new(seed, np)$state())
+})
```
johnlees, line 154: This is identical to the test immediately above

johnlees: Testing advance done below, so remove this one

tests/testthat/test-rng.R
```diff
@@ -265,3 +265,96 @@ test_that("get state", {
   expect_identical(s3[seq_len(32)], s1)
   expect_identical(s3[-seq_len(32)], rng2$jump()$state())
 })
+
+
+test_that("initialise single rng with binary state", {
+  seed <- 42
```
johnlees, line 271: If you do allow negative numbers, add a test here

tests/testthat/test-simulate.R
```diff
@@ -36,7 +36,7 @@ test_that("simulate multi-state model", {
   y0 <- matrix(c(1000, 10, 0), 3, np)
   steps <- seq(0, 200, by = 20)
 
-  ans <- dust_simulate(res, steps, data, y0)
+  ans <- dust_simulate(res, steps, data, y0, seed = 1L)
```
johnlees, line 39: Do we check that two calls to dust_simulate give different answers with `seed = NULL` somewhere?

inst/include/dust/rng_interface.hpp
richfitz: sure - I think I was mixing up the requirement for nonzero state in xoshiro with the requirements of the intermediate generator here

tests/testthat/test-simulate.R
richfitz: There's a check here: https://github.com/mrc-ide/dust/blob/i87-rng-seed/tests/testthat/test-rng.R#L323-L335

tests/testthat/test-rng.R
richfitz: done

inst/include/dust/rng_interface.hpp
richfitz: Done in 0b52ba3

tests/testthat/test-simulate.R
richfitz: oh, not for simulate specifically. I'll add that though as we mention it in the docs

richfitz: added in 3d44d55

inst/include/dust/interface.hpp
```diff
@@ -234,15 +234,13 @@ void dust_reorder(SEXP ptr, cpp11::sexp r_index) {
 }
 
 template <typename T>
-SEXP dust_rng_state(SEXP ptr, bool advance) {
+SEXP dust_rng_state(SEXP ptr, bool first_only) {
   Dust<T> *obj = cpp11::as_cpp<cpp11::external_pointer<Dust<T>>>(ptr).get();
   auto state = obj->rng_state();
-  size_t len = sizeof(uint64_t) * state.size();
+  size_t n = first_only ? dust::rng_state_t<double>::size() : state.size();
```
johnlees, line 240: template argument `double` or `T`?

johnlees: @richfitz I think this is right and will merge, but just a note if it needs changing later

PR #90, opened 2020-08-14
i89: Compile examples into package

### LINKED ISSUE #89: Add a compiled model into the package
https://github.com/mrc-ide/mcstate/issues/42

richfitz:
This PR bundles some of the examples into the package. This makes the compilation a little slower but has some advantages:

* we expose more of the code to coverage, which makes it easier to find code we've added and not checked
* we get faster examples, tests and vignettes because we don't pay the cost of compilation there (though it's cached)
* we get faster examples, tests and vignettes in mcstate because it can use the models here
* users do not need to configure a build environment to test-drive the package

Disadvantages:

* Package compilation speed, and therefore some of the dust development cycle is a bit slower
* The size of the libs directory has increased from ~600k to 4MB on Linux, less on macOS where I still see only 300k - this gives little headroom to avoid the 5MB `NOTE` in R CMD check, and including the cub headers in the package does push us over

Fixes #89 
Builds on #87 and should not be merged until after that PR


richfitz:
* `dust_file` is an internal convenience function and is not referenced in any user facing docs
* `dust::dust_package` has a bunch of checks and tests (tests/testthat/test-package.R, the "Use within a package" section of the vignette, and docs on `?dust::dust_package`)
* `dust::dust` is the old main way of running examples and is tested, and used in a number of remaining tests

PR #96, opened 2020-08-19
i94 openmp tweaks

### LINKED ISSUE #94: openmp pragmas should be behind ifdefs
To fully comply with R-exts etc, and prevent a compiler warning

### LINKED ISSUE #95: dust_package should validate Makevars
if src/Makevars does not contain the openmp boilerplate:

```
PKG_CXXFLAGS=$(SHLIB_OPENMP_CXXFLAGS)
PKG_LIBS=$(SHLIB_OPENMP_CXXFLAGS)
```

we should warn, and we might create this if it is not present

richfitz:
This PR closes two issues with openmp:

* we should have put all `#pragma omp...` usages behind an `#ifdef _OPENMP` check, which otherwises raises a compiler warning on linux when openmp is not used #94 
* we should check that a suitable src/Makevars is present to enable openmp or we end up with singlethreaded packages! #95 

Fixes #94 
Fixes #95


inst/include/dust/dust.hpp
```diff
@@ -139,30 +139,38 @@ class Dust {
   }
 
   void run(const size_t step_end) {
+#ifdef _OPENMP
```
johnlees, line 142: Here is a controversial thought: if you don't have an openmp compiler, how much is gained by using the package? Maybe a big fat warning when it doesn't understand the pragma would inspire a user to check the docs and try and enable it

johnlees (changes requested):

inst/include/dust/dust.hpp
richfitz: I don't have an openmp compiler on my laptop 😁 

The warnings would prevent other people writing models and getting them on CRAN, for example. Someone should be able to run sircovid on their non-openmp-enabled computer and get the same answer

johnlees (approved):

PR #99, opened 2020-08-19
i97: openmp info

### LINKED ISSUE #97: Report on openmp support
there are two issues here: 

1. can I compile with openmp?
2. can I run this model with openmp?

the former wants a free function in the package, the latter wants a method on the class I think

richfitz:
As discussed around #96, this PR starts building support for working out what sort of openmp support is available. It detects compilation support and individual models, and has support for picking sensible limits for the number of threads, following established conventions for signalling resource in process-level parallelism (e.g., future/furrr)

Fixes #97


tests/testthat/test-openmp.R
```diff
@@ -0,0 +1,145 @@
...
+  skip_if_not_installed("mockery")
+  mock_dust <- mockery::mock(
+    list(public_methods = list(has_openmp = function() TRUE)))
+  with_mock(
+    "dust::dust" = mock_dust,
+    expect_true(has_openmp_compiler_test()))
+})
+
+
+test_that("dust_openmp_threads inteface works", {
```
johnlees, line 132:
```suggestion
test_that("dust_openmp_threads interface works", {
```

johnlees: Proving that I do read these

R/openmp.R
```diff
@@ -0,0 +1,146 @@
...
+##'
+##' @title Select number of threads
+##'
+##' @param n Either `NULL` (select automatically) or an integer as
+##'   your proposed number of threads.
+##'
+##' @param action An action to perform if `n` exceeds the maximum
+##'   number of threads you can use. Options are "error" (the default,
+##'   throw an error), "fix" (print a message and reduce `n` down to
+##'   the limit) or "message" (print a message and continue anyway)
```
johnlees, line 62: This is a very 'feature rich' implementation. Not a problem, but I think many parallel implementations (e.g. openmp itself) would silently do the 'fix' behaviour.

tests/testthat/test-example.R
```diff
@@ -433,3 +433,11 @@ test_that("can run volatility example", {
   y <- drop(dust_iterate(obj, 0:100))
   expect_lt(diff(range(colMeans(y))), 0.5)
 })
+
+
+test_that("has_openmp can be called statically and normally", {
```
johnlees, line 438: Trying to work out whether `libgomp.so` is required - it may not be just using pragma (still unclear to me), but we use some of the runtime commands from the library (https://gcc.gnu.org/onlinedocs/libgomp/Runtime-Library-Routines.html#Runtime-Library-Routines) so I'm fairly sure it's required for us to run, when compiled with openmp

johnlees (changes requested):

R/openmp.R
richfitz: It will for the openmp part of the number, but not the R part, and I was thinking a bit of a message might be useful as it'd end up in a log (why did my thing not run any faster? I swear I increased the number of threads). And the CRAN issue looms as there are some fun issues over on data.table where they were told off by CRAN for not honouring some of these openmp limits

tests/testthat/test-example.R
richfitz: ah, good point

richfitz: ok, I've taken a better stab at this in 5c3f354

johnlees (approved):

PR #102, opened 2020-08-25
Add github actions

richfitz:
This PR swaps out appveyor/travis for github actions, and does the following things:

* Matrix build on windows/linux/mac for the package, doing R CMD check
* Simultaneous coverage check (rather than sequential)
* On merge to master, rebuild the pkgdown site (this might need tweaking as it's the default from `usethis` but it's hard to test without using it!)
* Create a release at every merge to master, by adding a tag and uploading binaries

The binaries are the main aim here really, as part of a system of providing binaries for packages that use C/C++ and are dependencies of other dept packages. But we'll probably start retiring appveyor/travis builds generally as this is where the wind is blowing!


.github/workflows/make-release.yaml
```diff
@@ -0,0 +1,155 @@
...
+          release_name: Release ${{ env.PACKAGE_NAME }} ${{ env.PACKAGE_VERSION }}
+          draft: false
+          prerelease: false
+
+      - name: Upload source
+        uses: actions/upload-release-asset@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          upload_url: ${{ steps.create_release.outputs.upload_url }}
```
r-ash, line 112: Where does this `upload_url` come from?

r-ash (approved):
This looks good, one question about where one of the variables is coming from.

.github/workflows/make-release.yaml
r-ash: Nice - I see that is in the output from `create_release` action

PR #104, opened 2020-10-08
Set rng state

richfitz:
This provides support for setting rng state into an existing dust model. I need this in order to support https://github.com/mrc-ide/mcstate/pull/79 as there we use this to separate out the model state and rng state components.


johnlees (approved):
This all looks good to me. Was a little confused by the two for loops over state, but fine after I refreshed my memory on how the rng object now looks

PR #107, opened 2020-10-29
i106: interleave rng

### LINKED ISSUE #106: Interleave and simplify rng state

richfitz:
This PR changes how we store RNG state, in preparation to work with the GPU version (again!). We now store rng state as a big single vector of `uint64_t` and pass to the model a struct that has a this data structure offset and strided. The blast-radius of this change is fairly small all things considered except for one breaking change:

* Models no longer accept `dust::rng_state_t<T>&` but instead `dust::rng_state_t<T>` (i.e., not a reference)

~This will require a little coordination when we're happy with this approach, namely:~

* ~Update odin.dust in to change the rng_state_t code to output the new type (https://github.com/mrc-ide/odin.dust/pull/31)~
* ~Regenerate sircovid~

(The above no longer true as the reference interface is back, though now optional)

No changes are needed in mcstate, nor in any of the code that *uses* the models, which is nice.

Previously we used a vector of objects for the RNG state, which was nice but played poorly on the GPU. It looks like we really want interleaved state so that for the rng state `[a, b, c, d]` if we had 3 particles we'd have `[a1, a2, a3, b1, b2, b3, c1, c2, c3, d1, d2, d3]`.

In the CPU version we allocate a big vector of `4 * n` `uint64_t` objects and as we get ready to use them we create a small object which holds a pointer to the `.data()` of this vector (a `*uint64_t` vector), offset by the particle index, and with a `operator[]` method that adjusts the index to jump the required number of elements ahead.

When we port to the GPU this means that we just have to do one big copy and I believe all the current CPU headers will just work, or be made to work

Fixes #106


inst/include/dust/rng.hpp
```diff
@@ -12,67 +12,82 @@ namespace dust {
...
     }
   }
 
-  rng_state_t<T>& state(size_t i) {
-    return _state[i];
+  // Access an individual rng_state by constructing the small struct
+  // that contains a pointer to the memory, offset as needed, and our
+  // stride.
+  rng_state_t<T> state(size_t i) {
+    return rng_state_t<T>(state_.data() + i, n_);
```
johnlees, line 57: Why doesn't this one have `i * n_` as below?

inst/include/dust/rng.hpp
```diff
@@ -12,67 +12,82 @@ namespace dust {
...
       if (i < n_seed) {
-        std::copy_n(seed.begin() + i * len, len, s.state.begin());
+        for (size_t j = 0; j < len; ++j) {
+          s[j] = seed[i * len + j];
+        }
       } else {
+        rng_state_t<T> prev = state(i - 1);
+        for (size_t j = 0; j < len; ++j) {
+          s[j] = prev[j];
+        }
```
johnlees, line 31: This isn't interleaved, is that right? Or is it coming from xoshiro where it is already interleaved?

inst/include/dust/rng.hpp
johnlees: If they are meant the be identical, could you use a define to set the functions as being the same? Or was that the fiddly bit you mentioned?

inst/include/dust/rng.hpp
```diff
@@ -12,67 +12,82 @@ namespace dust {
...
-    const size_t n = rng_state_t<T>::size();
-    state.reserve(size() * n);
-    for (size_t i = 0; i < size(); ++i) {
-      for (size_t j = 0; j < n; ++j) {
-        state.push_back(_state[i][j]);
+    const auto len = rng_state_t<T>::size();
+    std::vector<uint64_t> ret(n_ * len);
+    for (size_t i = 0; i < n_; ++i) {
+      for (size_t j = 0; j < len; ++j) {
+        ret[i * len + j] = state_[i + n_ * j];
```
johnlees, line 72: This de-interleaves? Can you note that in the comment above, I wasn't immediately sure what raw state would mean

inst/include/dust/xoshiro.hpp
```diff
@@ -39,7 +47,7 @@ static inline uint64_t rotl(const uint64_t x, int k) {
 
 // This is the core generator (next() in the original C code)
 template <typename T>
-inline uint64_t xoshiro_next(rng_state_t<T>& state) {
+inline uint64_t xoshiro_next(rng_state_t<T> state) {
```
johnlees, line 50: Looking at the above I am less sure why this shouldn't be a reference. Did you say that there was a compiler error?
I guess it's a fairly minor copy anyway, as the object is small

johnlees (changes requested):
Just a few things I wanted to clarify in the comments.

Also, I guess that a small update to odin.dust will be necessary to change the RNG state references to copies?

inst/include/dust/rng.hpp
richfitz: The other one was vestigial - I've removed it and if the tests pass that confirms it...

inst/include/dust/rng.hpp
richfitz: Added a comment now to clarify. You are right, it's not interleaved and this is only ever initialised from R

inst/include/dust/rng.hpp
richfitz: Added a comment

richfitz (commented):
For the comment

> Looking at the above I am less sure why this shouldn't be a reference. Did you say that there was a compiler error?
> I guess it's a fairly minor copy anyway, as the object is small

Now on deleted code, I looked further and the issue is the following. We create the `dust::rng_dust_t` object from the pRNG and *must* return a copy at that point; that is passed into `run` (interface.hpp) and so I've set things up now so that is the only place that accepts a non-reference object. All the others have been updated to references, which has lead to a bit of explicit object creation.

johnlees (approved):

PR #110, opened 2020-11-05
Temporarily remove gpu support

richfitz:
This version of the GPU support is too different to our new idea, so it makes more sense to remove entirely and rebuild. Here we had assumed that a dust object would be all-cpu or all-gpu, but in the new version it will be all-cpu with the *option* to run on a gpu, more like how we have the option to run with openmp. Accordingly, most of the user-facing gpu code goes away, and with the interleaved rng in #107 the separate stack of includes also goes away.


inst/include/dust/dust.hpp
```diff
@@ -4,6 +4,7 @@
 #include <dust/rng.hpp>
 
 #include <algorithm>
+#include <stdexcept>
```
richfitz, line 7: This is in an attempt to fix the failing release build on windows oldrel

johnlees (approved):

Issue #98, opened 2020-08-19
Vignette needs updating


richfitz:
In changing the seed behaviour in #88 we broke the breakage in the vignette: https://mrc-ide.github.io/dust/articles/dust.html#reset-the-model

johnlees:
Other collected thoughts:

- [ ] Noting that no underscores can be used in file/model names
- [ ] How to enable openmp on OS X (either through homebrew and gcc, or `brew install libomp` and compiler flags `-Xpreprocessor -fopenmp -lomp`)

PR #112, opened 2020-11-06
i108: coef() interface for dust

### LINKED ISSUE #108: Support for coef()

richfitz:
This PR restores an odin-like `coef()` interface for dust models. But because dust models could theoretically be written by hand, it's hard to control how these will be returned. 

What I did not want to do was embed some horrific metadata format within the C++ source code. So instead, I take advantage of `decor`, which we already depend on for use with `cpp11` and look for C++ (pseudo) attributes of the form

```
// [[dust::param(gamma, required = TRUE, default = 10)]]
```

(the format is arbitrary, and key-value combinations are allowed after the first argument which is the name).

This also fixes a long-standing wart, in that one could not change the classname or model name in the package interface, via attributes `[[dust::class(classname)]]` and `[[dust::name(modelname)]]`. These are not documented yet in the vignette, but I'd rather do that in a big documentation push (#98).

Returning a data.frame of results is hard because it requires every parameter to have the same fields, and these all be *always* atomic (this is not the case with odin, where the default requires a list column). So instead here we just return a tree of data as a list - see the examples in sir.cpp and sirs.cpp

In order to support `coef` on the generator objects I've added a class attribute at generation - this is good idea anyway IMO.

One outstanding question is if `param` is the best name for this, and not `coef`, `data`, `user` or something else

Fixes #108


richfitz:
The CodeFactor issue will be fixed on merge as it is a complaint about the length of a line in a generated file (R/dust.R), which has been ignored now on the CodeFactor side. This seems to only take effect after the merge, though if you click through no error will show

R/interface.R
```diff
@@ -79,18 +79,26 @@
 ##'   variable ordering, or any processing done while accepting the
 ##'   data object used to create the data fed into the particles.
 ##'
-##' @title Create a dust model from a C++ input file
+##' @section Confuring your model:
```
johnlees, line 82:
```suggestion
##' @section Configuring your model:
```

R/interface.R
```diff
@@ -79,18 +79,26 @@
...
+##'   use this if your file uses more than a single class, as
+##'   otherwise will try to detect this using extremely simple
+##'   heuristics.
 ##'
-##' @param type The name of the "type" (the C++ class) that represents
-##'   your model.  If \code{NULL} we try to work this out from your
-##'   file using extremely simple heuristics.
+##' * `[[dust::name(modelname)]]` will tell dust the name of the model
+##'   for exporting to R. For technical reasons this must be
+##'   alphanumeric characters only (sorry, no underscore) and must not
```
johnlees, line 95: Don't need to do this now, but in either #98 or odin.dust docs I think it's worth explicitly stating that underscores aren't allowed (as you do here). Particularly in model filenames, as I think that will be a common issue!

R/interface.R
```diff
@@ -79,18 +79,26 @@
...
 ##'   data object used to create the data fed into the particles.
 ##'
-##' @title Create a dust model from a C++ input file
+##' @section Confuring your model:
 ##'
-##' @param filename The path to a single C++ file
+##' You can optionally use C++ psuedo-attributes to configure the
+##'   generated code. Currently we support two attributes:
+##'
+##' * `[[dust::class(classname)]]` will tell dust the name of your
```
johnlees, line 87: i.e. the name of the class in the C++ code?

R/interface.R
```diff
@@ -79,18 +79,26 @@
...
+##'   target C++ class (in this example `classname`). You will need to
+##'   use this if your file uses more than a single class, as
+##'   otherwise will try to detect this using extremely simple
+##'   heuristics.
 ##'
-##' @param type The name of the "type" (the C++ class) that represents
-##'   your model.  If \code{NULL} we try to work this out from your
-##'   file using extremely simple heuristics.
+##' * `[[dust::name(modelname)]]` will tell dust the name of the model
```
johnlees, line 93: i.e. the name of the class in the R code?

R/metadata.R
```diff
@@ -0,0 +1,124 @@
+parse_metadata <- function(filename) {
+  data <- decor::cpp_decorations(files = filename)
```
johnlees, line 2: This overall approach seems very good to me

R/metadata.R
```diff
@@ -0,0 +1,124 @@
...
+    stop(sprintf("More than one [[%s()]] attribute found %s",
+                 attribute, parse_metadata_describe(data, i)))
+  }
+  value <- data$params[[which(i)]]
+  if (length(value) != 1L) {
+    stop(sprintf("Expected [[%s()]] to have one argument %s",
+                 attribute, parse_metadata_describe(data, i)))
+  }
+  if (any(nzchar(names(value)))) {
+    stop(sprintf("Invalid format for [[%s()]] attribute %s",
```
johnlees, line 36: Is this testing for empty strings? Could be a more informative error to add this, as format is quite broad

R/metadata.R
```diff
@@ -0,0 +1,124 @@
...
+parse_metadata_class <- function(data) {
+  parse_metadata_simple(data, "dust::class")
+}
+
+
+parse_metadata_param <- function(data) {
+  i <- data$decoration == "dust::param"
+  if (!any(i)) {
+    return(NULL)
+  }
```
johnlees, line 57: What is the behaviour when the decorators are not used for a parameter? Still compiles ok, but is `coef()` uncallable/incomplete?

R/metadata.R
```diff
@@ -0,0 +1,124 @@
...
+  if (any(!nzchar(names(x)[-1]))) {
+    stop(sprintf(
+      "Arguments 2 and following of of [[dust::param]] must be named %s",
+      parse_metadata_describe(data, i)),
+      call. = FALSE)
+  }
+
+  ## I think that we should allow only a restricted set here perhaps?
+  ## Or a general set special case some like required/default
```
johnlees, line 94: I don't follow this comment

R/utils.R
```diff
@@ -71,3 +71,8 @@ dust_header <- function(comment) {
   sprintf("%s Generated by dust (version %s) - do not edit",
           comment, utils::packageVersion("dust"))
 }
+
+
+squote <- function(x) {
+  sprintf("'%s'", x)
+}
```
johnlees, line 78: Do you have a dictionary of all of these!?
I tried to make an abind myself recently and it took a while

inst/examples/sir.cpp
```diff
@@ -62,7 +62,9 @@ sir::init_t dust_data<sir>(cpp11::list data) {
   double R0 = 0.0;
 
   // Rates, which can be set based on the provided data
+  // [[dust::param(beta, required = FALSE)]]
```
johnlees, line 65: This model doesn't also have the dust::model and dust::class descriptors, and they are still picked up automatically? Do we have a test somewhere for those (and that they can be different)?

tests/testthat/test-metadata.R
```diff
@@ -0,0 +1,181 @@
...
+test_that("Can override explicit class name with [[dust::name()]]", {
+  tmp <- helper_metadata(
+    "class someotherclass {};",
+    "// [[dust::class(walk)]]",
+    "// [[dust::name(model)]]")
+  on.exit(unlink(tmp))
+  meta <- parse_metadata(tmp)
+  expect_equal(meta$name, "model")
+  expect_equal(meta$class, "walk")
+})
```
johnlees, line 45: Ok great, I see that test here!

johnlees (changes requested):
Overall approach is great, and tests look thorough! A few questions in the comments

R/interface.R
richfitz: There will be an odin.dust update following this, so I can do that there

R/metadata.R
richfitz: yes, it keeps the current behaviour, and `coef()` etc return `NULL`

R/metadata.R
richfitz: agreed, updated with something better

R/metadata.R
richfitz: This was me wondering if the `param` attribute should have a restricted set of types here, of if it should treat some types specially - this is to help corral the returned data into a data.frame. However, I think that for now we'll forego that and I have removed the comment

R/utils.R
richfitz: This one I just write out as I need it. `abind` I'd generally use the package, but for our needs in sircovid it was simpler and likely faster to write one out. If it's a complicated utility I usually have a look for the last time I might have written it and copy it over from another package :)

johnlees (approved):

PR #116, opened 2020-11-30
Implement set_data method for replacing parameters

richfitz:
This is part of a (fairly ineffective) approach to speed up the sircovid vaccination work before we move the vaccination logic into the odin code. It provides a setter for `data` which swaps out the model parameters while holding everything else (state, index, etc) constant. It's not urgent to merge now but feels like a potentially useful thing to have.


R/dust_class.R
```diff
@@ -124,6 +124,16 @@ dust_class <- R6::R6Class(
     reset = function(data, step) {
     },
 
+    ##' @description
+    ##' Set the 'data' element in a dust object while holding parameters,
```
johnlees, line 128: I think 'parameters' here is a little unclear, as doesn't data often refer to what we might call parameters? Does parameters mean number of particles, threads, index and state? And this doesn't change state, correct?

tests/testthat/test-interface.R
```diff
@@ -254,3 +254,42 @@ test_that("Can get parameters from generators", {
...
+  res <- dust_example("variable")
+
+  obj <- res$new(list(len = 5, mean = 0, sd = 1), 0, 10, seed = 1L)
+  y1 <- obj$run(1)
+
+  expect_error(
+    obj$set_data(list(len = 6, mean = 10, sd = 10)),
+    paste("Tried to initialise a particle with a different size:",
+          "particle 1 had length 5 but data implies size 1"))
+  expect_identical(obj$state(), y1)
```
johnlees, line 289: Still not totally sure what 'length' is in the error here. Is it if you change some of the internal variable's dimensions?

johnlees (changes requested):

R/dust_class.R
richfitz: parameters here should have been model state

tests/testthat/test-interface.R
richfitz: yeah, this is unclear. I've updated the docs

johnlees (approved):

PR #120, opened 2020-12-10
Optionally return model end state with dust_simulate

richfitz:
This is needed to support picking the simulation back up, which is how we will do time-varying vaccination simulation in sircovid.


inst/include/dust/dust.hpp
```diff
@@ -290,10 +290,11 @@ template <typename T>
 std::vector<typename T::real_t>
 dust_simulate(const std::vector<size_t> steps,
               const std::vector<typename T::init_t> data,
-              const std::vector<typename T::real_t> state,
+              std::vector<typename T::real_t>& state,
```
johnlees, line 293: Should data (on the line above) also be passed by reference. I got stung by some of these being missing when I scaled up to 100k particles, causing endless copies of enormous vectors!

inst/include/dust/interface.hpp
```diff
@@ -79,12 +80,21 @@ cpp11::writable::doubles dust_simulate(cpp11::sexp r_steps,
...
+    dust_simulate<T>(steps, data, state, index, n_threads, seed, return_state);
   std::copy(dat.begin(), dat.end(), REAL(ret));
 
   ret.attr("dim") = cpp11::writable::integers({(int)index.size(),
                                                (int)data.size(),
                                                (int)steps.size()});
+
+  if (return_state) {
+    cpp11::writable::doubles r_state_end(state.size());
+    std::copy(state.begin(), state.end(), REAL(r_state_end));
```
johnlees, line 92: No way around the two copies due to conflicting std::vector and cpp11 vector types?

johnlees (changes requested):

inst/include/dust/dust.hpp
richfitz: good shout

inst/include/dust/interface.hpp
richfitz: exactly - we'd have to template much of the underlying functions and make sure we had a proxy class that made the REALSXP behave enough like an iterator - it could be done but not in this PR :)

johnlees (approved):

PR #123, opened 2020-12-23
New method to change the number of threads (#122)

### LINKED ISSUE #122: Change number of threads without reinitialisation

richfitz:
Allows changing the number of threads in an already-initialised dust object. This is needed so that we can dynamically scale the number of cores used for the pmcmc during a run in mcstate.

Fixes #122

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: dust
 Title: Iterate Multiple Realisations of Stochastic Models
-Version: 0.5.6
+Version: 0.5.7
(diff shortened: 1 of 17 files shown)


johnlees (approved):
Should we add a message/error if this is used when openmp isn't available? (perhaps the same if initialised with n_threads > 1 too)

Otherwise looks good. Happy to merge if there's a reason you don't want to add the above

richfitz:
This is handled via `dust_openmp_threads` which the user can wrap the argument in. I've added a pointer to this in the docs in 6ae2c57

PR #124, opened 2020-12-24
Example demonstrating shared constant memory

### LINKED ISSUE #118: Support shared constant memory within a group of dust objects
As discussed with John, and useful for the gpu, and may help keep openmp performance linear in the number of threads?

richfitz:
I think it's preferable to start this one purely as a pattern and not complicate the interface much further.

~Merge after #123 as it builds on it~
Fixes #118

### DIFF
--- a/inst/examples/shared.cpp
+++ b/inst/examples/shared.cpp
@@ -68,23 +68,22 @@ private:
 
 template<>
 arrays::init_t dust_data<arrays>(cpp11::list user) {
-  arrays::shared_t shared;
-  shared.initial_y = 2;
-  shared.n = 3;
-  shared.dim_r = shared.n;
-  shared.dim_x = shared.n;
-  for (int i = 0; i < shared.dim_x; ++i) {
-    shared.initial_x[i] = 1;
+  auto shared = std::make_shared<arrays::shared_t>();
+
+  shared->initial_y = 2;
+  shared->n = 3;
+  shared->dim_r = shared->n;
+  shared->dim_x = shared->n;
+  for (int i = 0; i < shared->dim_x; ++i) {
+    shared->initial_x[i] = 1;
   }
 
   arrays::internal_t internal;
-  for (int i = 0; i < shared.dim_r; ++i) {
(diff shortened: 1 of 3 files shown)


inst/examples/shared.cpp
```diff
@@ -0,0 +1,97 @@
...
+public:
+  typedef double real_t;
+
+  struct shared_t {
+    int dim_r;
+    int dim_x;
+    int n;
+    real_t initial_y;
+    std::array<real_t, 3> initial_x;
+  };
```
johnlees, line 23: Do we definitely only hold one copy of this in memory, as every particle will contain an `arrays` class, each of which has a shared_t struct? Should this be `static`?
Or am I misunderstanding?

johnlees (changes requested):

inst/examples/shared.cpp
richfitz: I must admit I never remember all the different ways that `static` can be used. But if it's for a struct or class member would that not shared by all instantiations of the class? Or are you meaning something else? (Where in the above are you thinking of putting `static`?)

johnlees: i.e.:
```
  static struct shared_t {
    int dim_r;
    int dim_x;
    int n;
    real_t initial_y;
    std::array<real_t, 3> initial_x;
  };
```
So that it is shared by all instantiations of the class - isn't that what we want?

richfitz: It's not though - if two copies of this are created with different `n` then they'll have different values. It's the same for all particles within a dust object though

johnlees: Right, yes, that makes sense - no static then.
I'm getting in a muddle between struct definition and instance I think. `dust_data` actually creates this struct and the pointer, and is only called once?

richfitz: yeah, this is just the definition of the struct which basically namespaces it for us

johnlees: Right, I will merge then!

johnlees (approved):

PR #128, opened 2021-01-12
Implement rexp

### LINKED ISSUE #127: Add exponential distribution
Will be needed to implement the compare function for sircovid

richfitz:
Basic implementation of exponential random samples. Requires log() but at least it's really easy to understand.

Fixes #127

### DIFF
--- a/R/dust_class.R
+++ b/R/dust_class.R
@@ -227,7 +227,7 @@ dust_class <- R6::R6Class(
     ##' @description
     ##' Returns the number of distinct data elements required. This is `0`
     ##' where the object was initialised with `data_multi = FALSE` and
-    ##' an integer otherwise.  For multi-data dust objects, Where `data`
+    ##' an integer otherwise.  For mulit-data dust objects, Where `data`
     ##' is accepted, you must provide an unnamed list of length `$n_data()`.
     n_data = function() {
     },
--- a/R/dust_class.R
+++ b/R/dust_class.R
@@ -227,7 +227,7 @@ dust_class <- R6::R6Class(
     ##' @description
     ##' Returns the number of distinct data elements required. This is `0`
     ##' where the object was initialised with `data_multi = FALSE` and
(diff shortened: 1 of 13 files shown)


johnlees:
Think this makes sense for now. Just noting the R source if we want to look at inversion in future:
https://github.com/wch/r-source/blob/d22ee2fc0dc8142b23eed9f46edf76ea9d3ca69a/src/nmath/sexp.c

R/dust_class.R
```diff
@@ -227,7 +227,7 @@ dust_class <- R6::R6Class(
     ##' @description
     ##' Returns the number of distinct data elements required. This is `0`
     ##' where the object was initialised with `data_multi = FALSE` and
-    ##' an integer otherwise.  For multi-data dust objects, Where `data`
+    ##' an integer otherwise.  For mulit-data dust objects, Where `data`
```
johnlees, line 230:
```suggestion
    ##' an integer otherwise.  For multi-data dust objects, Where `data`
```
Merge mistake?

inst/include/dust/distr/exponential.hpp
```diff
@@ -0,0 +1,31 @@
...
+// Chapter 9, p 392
+//
+// > No method is shorter than the inversion method, which returns
+// > -log(U) where U is a uniform [0,1] random variate
+//
+// Faster generators will exist but we can swap one in if it becomes
+// important.
+template <typename real_t>
+real_t exp_rand(rng_state_t<real_t>& rng_state) {
+  return -log(dust::unif_rand(rng_state));
```
johnlees, line 20: `std::log`? That's what we use in rbinom (think it might be to help with CUDA as it works in both case, but I'm not actually sure of the reason)

johnlees (changes requested):

R/dust_class.R
richfitz: This file is generated from one elsewhere, will track the source down

johnlees (approved):

PR #132, opened 2021-01-14
Basic implementation of compare in dust

### LINKED ISSUE #129: Support for compiled compare functions
Lots of work needed here to support this, including upstream in odin.dust and possibly odin to prepare use with sircovid.

* Optionally store some data object within the dust object. We might want to update the name here to pars rather than data as this is getting confusing. This should be updated less frequently than the parameters (data).
* Add to the particles a compare function like `.compare(rng_state, step, data)` which will run the compare for the current step

To get the step->data map we could store the actual data as `std::map<size_t, T>` so that we get nice fast lookup of the data given step, though that means that if the model varies dt then we have to deal with this by resetting data. Alternatively we could allow `.compare(rng_state, data)` and assume that the model is tracking its *own* time

We also need to rewrite the compare approach in sircovid and mcstate to not use the previous date as a way of computing incidence but that's a good thing to tidy up anyway

Make sure that additional but unused user parameters make it into the parameters object as we'll need that for things like exp_noise

Previously #55

richfitz:
This requires a bit of work to be done in odin.dust and it would be good to get a proof-of-concept working in mcstate too, but otherwise I think this is shaping up nicely.

The biggest ick I have so far is that I can't make the `data_t` optional because my SFINAE/C++ metaprogramming skills are pretty basic. A much simpler version of the problem is below with the aim to get the `null_type` there made optional so that `B` does not have to hold it, but still allow `C<B>` to be constructed.

<details>
<summary>Details</summary>

```c++
#include <map>
#include <iostream>

struct null_type {
};

class A {
public:
  typedef double data_t;
  double f(double a) {
    return 1;
  }
};

class B {
public:
  typedef null_type data_t;
};

template <typename T>
class C {
public:
  typedef typename T::data_t data_t;
  std::map<size_t, data_t> data;
};

template <typename T, typename std::enable_if<!std::is_same<null_type, typename T::data_t>::value, int>::type = 0>
double f(T x) {
  return x.f(1);
}

template <typename T, typename std::enable_if<std::is_same<null_type, typename T::data_t>::value, int>::type = 0>
double f(T x) {
  return 0;
}

int main() {
  C<A> a;
  C<B> b;
  double x = f(A());
  double y = f(B());
  std::cout << "x: " << x << std::endl;
  std::cout << "y: " << y << std::endl;
 
  return 0;
}
```
</details>

Fixes #129


inst/examples/sir2.cpp
```diff
@@ -0,0 +1,123 @@
...
+inline double with_default(double default_value, cpp11::sexp value) {
+  return value == R_NilValue ? default_value : cpp11::as_cpp<double>(value);
+}
+
+template <>
+sir2::init_t dust_pars<sir2>(cpp11::list pars) {
+  // Initial state values
+  double I0 = 10.0;
+  double S0 = 1000.0;
+  double R0 = 0.0;
```
johnlees, line 85: `real_t` template?

inst/examples/sir2.cpp
```diff
@@ -0,0 +1,123 @@
...
+    const double incidence_modelled = state[4];
+    const double incidence_observed = data.incidence;
+    const double lambda = incidence_modelled +
+      dust::distr::rexp(rng_state, pars_.exp_noise);
+    return ldpois(incidence_observed, lambda);
+  }
+
+private:
+  init_t pars_;
+};
```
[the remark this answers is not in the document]
johnlees: Also I think the doubles should be `real_t`

inst/examples/sir2.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+  // Rates, which can be set based on the provided pars
+  // [[dust::param(beta, required = FALSE)]]
+  double beta = with_default(0.2, pars["beta"]);
+  // [[dust::param(gamma, required = FALSE)]]
+  double gamma = with_default(0.1, pars["gamma"]);
+
+  // Time scaling
+  size_t freq = 4;
+  double dt = 1.0 / static_cast<double>(freq);
```
johnlees, line 95: As we write more of these models, it increasingly feels like an integer steps per data is the thing we want to specify rather than dt. Is there a way we can enforce/formalise this (perhaps in mcstate) or is it too model-dependent?

inst/examples/variable.cpp
```diff
@@ -1,6 +1,7 @@
 class variable {
 public:
   typedef double real_t;
+  typedef dust::no_data data_t;
```
johnlees, line 4:
```suggestion
  typedef dust::no_data data_t;
  
```

inst/examples/vectors.cpp
```diff
@@ -3,6 +3,7 @@
 class vectors {
 public:
   typedef double real_t;
+  typedef dust::no_data data_t;
```
johnlees, line 6:
```suggestion
  typedef dust::no_data data_t;
  
```

inst/examples/volatility.cpp
```diff
@@ -1,6 +1,7 @@
 class volatility {
 public:
   typedef double real_t;
+  typedef dust::no_data data_t;
```
johnlees, line 4:
```suggestion
  typedef dust::no_data data_t;
  
```

inst/examples/walk.cpp
```diff
@@ -1,6 +1,7 @@
 class walk {
 public:
   typedef double real_t;
+  typedef dust::no_data data_t;
```
johnlees, line 4:
```suggestion
  typedef dust::no_data data_t;
  
```

inst/include/dust/dust.hpp
```diff
@@ -81,6 +88,11 @@ class Particle {
     }
   }
 
+  double compare_data(const data_t& data,
```
johnlees, line 91: `real_t`?

inst/include/dust/dust.hpp
```diff
@@ -286,11 +299,31 @@ class Dust {
     _rng.long_jump();
   }
 
+  void set_data(std::map<size_t, data_t> data) {
+    _data = data;
+  }
+
+  std::vector<double> compare_data() {
+    std::vector<double> res;
```
johnlees, line 307: `real_t`

inst/include/dust/dust.hpp
```diff
@@ -4,18 +4,25 @@
 #include <dust/rng.hpp>
 
 #include <algorithm>
+#include <map>
```
johnlees, line 7: As far as I can tell, we always access individual elements of the `map`, and don't iterate through it. So I think we should be using `unordered_map`.

inst/include/dust/dust.hpp
```diff
@@ -286,11 +299,31 @@ class Dust {
...
   }
 
+  void set_data(std::map<size_t, data_t> data) {
+    _data = data;
+  }
+
+  std::vector<double> compare_data() {
+    std::vector<double> res;
+    auto d = _data.find(step());
+    if (d != _data.end()) {
```
johnlees, line 309: What about the else case here? Does the interface turn this uninitialised vector into a null for us?

inst/include/dust/interface.hpp
```diff
@@ -2,6 +2,7 @@
 #define DUST_INTERFACE_HPP
 
 #include <cstring>
+#include <map>
```
johnlees, line 5: ditto `map` -> `unordered_map`

inst/include/dust/interface.hpp
```diff
@@ -406,6 +410,56 @@ cpp11::sexp dust_info(const typename T::init_t& pars) {
...
+template <typename T, typename std::enable_if<std::is_same<dust::no_data, typename T::data_t>::value, int>::type = 0>
+void dust_set_data(SEXP ptr, cpp11::list r_data) {
+  cpp11::stop("The 'set_data' method is not supported for this class");
+}
+
+template <typename T, typename std::enable_if<std::is_same<dust::no_data, typename T::data_t>::value, int>::type = 0>
+cpp11::sexp dust_compare_data(SEXP ptr) {
+  cpp11::stop("The 'compare_data' method is not supported for this class");
+  return R_NilValue; // never gets here
+}
```
johnlees, line 449: Perhaps just make a comment above these classes to note that the template metaprogramming leads to compilation of this function if data is null/missing; the full function above if instead it's given

inst/include/dust/interface.hpp
```diff
@@ -406,6 +410,56 @@ cpp11::sexp dust_info(const typename T::init_t& pars) {
...
+    const data_t data_i = dust_data<T>(cpp11::as_cpp<cpp11::list>(el[1]));
+    data[step_i] = data_i;
+  }
+  obj->set_data(data);
+}
+
+template <typename T, typename std::enable_if<!std::is_same<dust::no_data, typename T::data_t>::value, int>::type = 0>
+cpp11::sexp dust_compare_data(SEXP ptr) {
+  Dust<T> *obj = cpp11::as_cpp<cpp11::external_pointer<Dust<T>>>(ptr).get();
+  std::vector<double> ret = obj->compare_data();
```
johnlees, line 431: `real_t`

vignettes/dust.Rmd
```diff
@@ -47,6 +47,7 @@ There are two main parts here; the class (`walk`) and an interface function (`du
 The class is what does most of the work. It is important that this does not use anything in the R api as it may be called in parallel. Therefore use only C++ standard library types.  It has to provide every public type or method as the example above:
 
 * the type `real_t` has to exist and indicate what sort of numeric type you use for "real numbers". Typically this will be `double`,but if you need a different type (longer or shorter) you can use it here.
+* the type `data_t` which for now should match exactly the construction `typedef dust::no_data data_t` until we stabilise and document this interface
```
johnlees, line 50: Is there anywhere in the vignette where `has_openmp` needs changing? Or did this remain unchanged as there was still an R function which digs into capabilities and checks this, with the same function name as before

johnlees (changes requested):
Templating and overall approach both look sensible to me, and ends up being fairly easy to follow.

A few specific comments, and also can you double-check all uses of `double`: some should probably be `real_t`.

inst/examples/sir2.cpp
```diff
@@ -0,0 +1,123 @@
+double ldpois(double x, double lambda) {
+  return x * std::log(lambda) - lambda - std::lgamma(x + 1);
+}
```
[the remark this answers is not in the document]
richfitz: This is not package code, just for the example.

I'd hope that anyone writing a likelihood is ok to do basic log transformations but you may be right.

I think that we probably will end up with a family of useful functions in dust that can be included (themselves using a log = true / log = false argument or a ld* and d* form), unless we can find a lightweight header-only set of these with tuneable precision

inst/examples/sir2.cpp
```diff
@@ -0,0 +1,123 @@
...
+    double gamma;
+    double dt;
+    size_t freq;
+    // Observation parameters
+    double exp_noise;
+  };
+  // Presence of this is going to be important
+  struct data_t {
+    double incidence;
+  };
```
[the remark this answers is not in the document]
richfitz: It's not static because it ends up in `std::map<size_t, data_t>` with each timestep having a different `data_t` with a different incidence value

This example could have been simplified further by doing `typedef double data_t` but I wanted to show creating a struct here as that's the most likely solution

inst/examples/sir2.cpp
```diff
@@ -0,0 +1,123 @@
...
+  struct init_t {
+    double S0;
+    double I0;
+    double R0;
+    double beta;
+    double gamma;
+    double dt;
+    size_t freq;
+    // Observation parameters
+    double exp_noise;
```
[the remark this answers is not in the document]
richfitz: Yes in odin.dust that will all end up in shared_t, but that's not a dust concept

inst/examples/sir2.cpp
```diff
@@ -0,0 +1,123 @@
...
+    double n_IR = dust::distr::rbinom(rng_state, I, p_IR * pars_.dt);
+    double n_SI = dust::distr::rbinom(rng_state, S, p_SI * pars_.dt);
+
+    state_next[0] = S - n_SI;
+    state_next[1] = I + n_SI - n_IR;
+    state_next[2] = R + n_IR;
+    state_next[3] = cumulative_incidence + n_SI;
+    // Little trick here to compute daily incidence by accumulating
+    // incidence from the first day.
+    state_next[4] = (step % pars_.freq == 0) ? n_SI : state[4] + n_SI;
```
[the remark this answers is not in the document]
richfitz: `dt => 1/freq`, see line 94-95

R/dust_class.R
```diff
@@ -242,6 +242,37 @@ dust_class <- R6::R6Class(
...
+    },
+
+    ##' @description
+    ##' Compare the current model state against the data as set by
+    ##' `set_data`. If there is no data set, or no data corresponding to
+    ##' the current time then `NULL` is returned. Otherwise a numeric vector
+    ##' the same length as the number of particles is returned. If model's
+    ##' underlying `compare_data` function is stochastic, then each call to
+    ##' this function may be result in a different answer.
+    compare_data = function() {
```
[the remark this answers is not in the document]
richfitz: good point, was being agnostic but in mcstate we do rely on that

inst/examples/sir2.cpp
```diff
@@ -0,0 +1,123 @@
...
+private:
+  init_t pars_;
+};
+
+#include <cpp11/list.hpp>
+
+// Helper function for accepting values with defaults
+inline double with_default(double default_value, cpp11::sexp value) {
+  return value == R_NilValue ? default_value : cpp11::as_cpp<double>(value);
+}
```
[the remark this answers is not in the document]
richfitz: yeah, odin.dust does this in a much more sophisticated way

inst/examples/sir2.cpp
richfitz: We do have this concept in mcstate already so yes. However, there are some models that don't use that sort of scaling because they have a strong sense of time (wright-fisher etc)

inst/include/dust/dust.hpp
richfitz: We get a vector of length 0, which is tested against in the interface and returns `NULL` back to R - that ends up being the same behaviour as the mcstate R compare version

richfitz: However, that's not at all obvious so I will add a comment :)

inst/include/dust/dust.hpp
richfitz: yup, that was incorrect

vignettes/dust.Rmd
richfitz: yeah, the public interface is unchanged

inst/examples/sir2.cpp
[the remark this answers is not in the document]
richfitz: as you found later, data_t is held by dust and I've updated the doubles to real_t (we should update the others too: #133)

inst/examples/sir2.cpp
richfitz: Also, we generally need access to `dt` from the update function but the parameter type is opaque to dust so if we make it part of the interface then it's not obvious how we get it to the update and compare function without complicating their call signature. This is worth thinking about though

johnlees (approved):

PR #135, opened 2021-01-18
Support for basic density functions

### LINKED ISSUE #134: Add library of compare functions
As with the 'r' functions (e.g. `rbinom`, `rexp` etc) it may be useful to provide out-of-the-box 'd' versions which give the density in the comparison function. We could either do the same functions as we have random deviates for, or target some known epi models specifically.

We should also probably deal with log explicitly (either `log = TRUE/FALSE` or make `dpois` and `ldpois`)

richfitz:
Added support just for the functions that we need in sircovid for now

Fixes #134


tests/testthat/test-densities.R
```diff
@@ -0,0 +1,57 @@
+context("densities")
+
+## This one is the easiest:
+test_that("dpois agrees", {
```
johnlees, line 4: I think we should add an edge case test to all of these:
n = 0; p/lambda = 0; p = 1 are the ones I can think of (and perhaps combinations of these)

`dpois(1, 0, log = TRUE) = -Inf` is the kind of thing that motivated `rexp`

inst/include/dust/densities.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+template <typename T>
+T lfactorial(T x) {
+  return std::lgamma(x + 1);
+}
+
+template <typename T>
+T lbeta(T x, T y) {
+  return lgamma(x) + lgamma(y) - lgamma(x + y);
+}
```
johnlees, line 28: I can't remember if we're C++14 or C++17, but if the latter this (unlogged) is in <cmath> as std::beta

inst/include/dust/densities.hpp
```diff
@@ -0,0 +1,73 @@
+#ifndef DUST_DENSITIES_HPP
+#define DUST_DENSITIES_HPP
+
+#include <cmath>
+
+namespace dust {
+
+template <typename T>
+T maybe_log(T x, bool log) {
```
johnlees, line 9: `inline` this I'd say?

johnlees: Possibly inline all the functions here

inst/include/dust/densities.hpp
```diff
@@ -0,0 +1,73 @@
...
+    size * std::log(prob) + x * std::log(1 - prob);
+  return maybe_log(ret, log);
+}
+
+// A note on this parametrisation:
+//
+//   prob = alpha / (alpha + beta)
+//   rho = 1 / (alpha + beta + 1)
+//
+// Where alpha and beta have (0, Inf) support
```
johnlees, line 56: I'll take your word for it 🙂

johnlees (changes requested):

inst/include/dust/densities.hpp
richfitz: We really need the logged version though, and that's not coming in either standard

inst/include/dust/densities.hpp
richfitz: All template functions are inline

inst/include/dust/densities.hpp
richfitz: Haha, this is from @edknock

inst/include/dust/densities.hpp
johnlees: 🤯

johnlees: I feel like I should have known that

richfitz: Seems to depend on the exact definition of "inline", per SO

inst/include/dust/densities.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+template <typename T>
+T dnbinom(int x, int size, T mu, bool log) {
+  const T prob = size / (size + mu);
+  if (x == 0 && size == 0) {
+    return maybe_log(0, log);
+  }
+  const T ret = std::lgamma(static_cast<T>(x + size)) -
+    std::lgamma(static_cast<T>(size)) -
+    lfactorial(static_cast<T>(x)) +
```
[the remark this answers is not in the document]
richfitz: I don't think it's that unreadable as maths :) But I have dropped the lfactorial as it's not getting us much

tests/testthat/test-densities.R
richfitz: good shout - added the cases I could think of here, can add more now or later as we find them

johnlees (approved):

PR #144, opened 2021-01-28
Add shared memory type to dust models

### LINKED ISSUE #133: Update examples to avoid double in favour of real_t

### LINKED ISSUE #143: Formalise constant shared memory approach
We're going to lean on this a bit with the GPU implementation

richfitz:
This is a disruptive breaking change, but required to support the GPU work. Updates are needed to odin.dust (and then to sircovid) to be merged in sync.

This simply formalises the approach that we are using as a pattern in the example (and odin.dust), and for a sufficiently simple struct will allow copying directly to the device (we might need to do some tricks where the structs contain `std::vector<real_t>`, but we can do that by an optional translation method I think)

Fixes #143 
Fixes #133 
Fixes https://github.com/mrc-ide/odin.dust/issues/58

Follow-up odin.dust PR: https://github.com/mrc-ide/odin.dust/pull/60
Follow-up sircovid PR: https://github.com/mrc-ide/sircovid/pull/207


inst/examples/arrays.cpp
```diff
@@ -5,63 +5,75 @@ class arrays {
...
-    for (int i = 0; i < internal.dim_x; ++i) {
+    state_next[0] = y + shared->n;
+    for (int i = 0; i < shared->dim_x; ++i) {
       state_next[i] = x[i] + internal.r[i];
     }
   }
 
 private:
-  init_t internal;
+  std::shared_ptr<const shared_t> shared;
```
johnlees, line 48: Why is this one `const`, but others aren't?

inst/examples/vectors.cpp
```diff
@@ -5,66 +5,75 @@ class vectors {
...
-    for (int i = 0; i < internal.dim_x; ++i) {
+    state_next[0] = y + shared->n;
+    for (int i = 0; i < shared->dim_x; ++i) {
       state_next[i] = x[i] + internal.r[i];
     }
   }
 
 private:
-  init_t internal;
+  std::shared_ptr<const shared_t> shared;
```
johnlees, line 45: This one also `const`

johnlees (changes requested):
I might be misunderstanding something again here, but if the model class has a shared pointer to shared_t, how does this work if we're doing multiple parameter sets in the dust object? Wouldn't this mean they are all forced to be the same single struct?

inst/examples/arrays.cpp
richfitz: Due to `using` in `dust.hpp` `dust::shared_ptr<T>` is the same thing as `std::shared_ptr<const typename typedef T::shared_t>` (or here `std::shared_ptr<const shared_t>`). These two files arrays.cpp and vectors.cpp were the first I ported and before I wrote that helper; I've moved them over as of ab25667

richfitz:
> I might be misunderstanding something again here, but if the model class has a shared pointer to shared_t, how does this work if we're doing multiple parameter sets in the dust object? Wouldn't this mean they are all forced to be the same single struct?

1. Create a vector of `pars_t` objects each with their own shared pointer: https://github.com/mrc-ide/dust/blob/i143-shared/inst/include/dust/interface.hpp#L59
2. Loop over this list of parameters when creating the particles: https://github.com/mrc-ide/dust/blob/i143-shared/inst/include/dust/dust.hpp#L383

inst/examples/arrays.cpp
richfitz: ugh, and then fixed in 71a7d8d

richfitz:
Totally baffled: the diff there is completely harmless but we're failing on all platforms and not locally for me daee40d...71a7d8d

johnlees:
> > I might be misunderstanding something again here, but if the model class has a shared pointer to shared_t, how does this work if we're doing multiple parameter sets in the dust object? Wouldn't this mean they are all forced to be the same single struct?
> 
> 1. Create a vector of `pars_t` objects each with their own shared pointer: https://github.com/mrc-ide/dust/blob/i143-shared/inst/include/dust/interface.hpp#L59
> 2. Loop over this list of parameters when creating the particles: https://github.com/mrc-ide/dust/blob/i143-shared/inst/include/dust/dust.hpp#L383

Ah right, thanks for reminding me... So when this becomes CUDA we can reach into the first particle to get the shared_t to copy

johnlees (approved):

richfitz:
Yeah, there will likely be more required there though if the shared_t includes things like std::vector (which it will for odin models at the moment) but we can cross that bridge when we get to it...

PR #151, opened 2021-02-08
Implement resampling

### LINKED ISSUE #136: compare should cope properly with multiple data

### LINKED ISSUE #137: Implement resampling

richfitz:
This does not depend on the model having a `compare_data` method as here we take a vector of weights. Uses the same algorithm as mcstate, though the implementations *look* very different.

This PR also fixes treatment of compare with multiple pars (#136) which now returns a matrix when using this mode.

Fixes #137 
Fixes #136


inst/include/dust/dust.hpp
```diff
@@ -282,15 +283,45 @@ class Dust {
...
-      p.swap();
+#ifdef _OPENMP
+    #pragma omp parallel for schedule(static) num_threads(_n_threads)
+#endif
+    for (size_t i = 0; i < _particles.size(); ++i) {
+      _particles[i].swap();
     }
   }
 
+  std::vector<size_t> resample(const std::vector<real_t>& weights) {
```
johnlees, line 301: Can we add a boolean `log` argument here so that this can work with log weights, where we add the scale_log_weights() function which deals with the log-sum-exp issue? I'm sure this function was going to be added when the full particle filter implementation appears soon, but seems like it might be handy to have this as an option here too

inst/include/dust/dust.hpp
```diff
@@ -282,15 +283,45 @@ class Dust {
...
+    std::vector<size_t> idx(n_particles());
+    auto it_weights = weights.begin();
+    auto it_idx = idx.begin();
+    if (_n_pars == 0) {
+      const size_t np = _particles.size();
+      real_t u = dust::unif_rand(_rng.state(0));
+      resample_weight(it_weights, np, u, 0, it_idx);
+    } else {
```
johnlees, line 309: Double-checking: these branches are for when there are multiple parameter sets, and shuffles only occur within the particle set? Might be helpful to add a comment to that effect

inst/include/dust/dust.hpp
```diff
@@ -282,15 +283,45 @@ class Dust {
...
     }
   }
 
+  std::vector<size_t> resample(const std::vector<real_t>& weights) {
+    std::vector<size_t> idx(n_particles());
+    auto it_weights = weights.begin();
+    auto it_idx = idx.begin();
+    if (_n_pars == 0) {
+      const size_t np = _particles.size();
+      real_t u = dust::unif_rand(_rng.state(0));
```
johnlees, line 307: Taking the first RNG state seems like it should be fine (and below) as we're unlikely to generate huge amounts of numbers here, and the number of times they're used between particles varies anyway.

I think there's also an argument for having a separate RNG for particle filter purposes: would avoid needing to copy this state to and from the device in the CUDA code (it would stay on the host only); keeps model simulations repeatable with more independence from how they have been run.
It would make the OPENMP block below less neat though.
In the CUDA, if we're doing it this way, I could also separate out the update_host/device functions to just do the RNG state.

What do you think?

inst/include/dust/interface.hpp
```diff
@@ -371,6 +371,53 @@ void dust_reorder(SEXP ptr, cpp11::sexp r_index) {
...
+SEXP dust_resample(SEXP ptr, cpp11::doubles r_weights) {
+  Dust<T> *obj = cpp11::as_cpp<cpp11::external_pointer<Dust<T>>>(ptr).get();
+  size_t n_particles = obj->n_particles();
+  size_t n_pars = obj->n_pars();
+  size_t n_particles_each = n_particles;
+  if (n_pars == 0) {
+    if (static_cast<size_t>(r_weights.size()) != n_particles) {
+      cpp11::stop("Expected a vector with %d elements for 'weights'",
+                  n_particles);
+    }
```
johnlees, line 384: Any other checks on weights needed?:

- Must they sum to 1?
- At least one weight > 0? 
- No negative weights?

inst/include/dust/tools.hpp
```diff
@@ -0,0 +1,27 @@
+#ifndef DUST_TOOLS_HPP
+#define DUST_TOOLS_HPP
+
+#include <algorithm>
+#include <numeric>
+
+template <typename real_t>
+void resample_weight(typename std::vector<real_t>::const_iterator w,
+                     size_t n, real_t u, size_t offset,
+                     typename std::vector<size_t>::iterator idx) {
```
johnlees, line 10: Correct for these iterators to be copies?

tests/testthat/test-example.R
```diff
@@ -505,3 +505,33 @@ test_that("fetch model size", {
...
+
+
+test_that("resample", {
+  res <- dust_example("variable")
+  obj <- res$new(list(len = 5), 0, 7, seed = 1L)
+  m <- matrix(as.numeric(1:35), 5, 7)
+  obj$set_state(m)
+  rng <- dust_rng$new(obj$rng_state(TRUE))
+  u <- rng$unif_rand(1)
+  w <- runif(obj$n_particles())
```
johnlees, line 517: Can we try a couple of boundary cases on weights? e.g. only one non-zero weight, all zero-weights (should error)

johnlees (changes requested):

inst/include/dust/tools.hpp
richfitz: yes, this is normal style I think for iterators (the iterator itself will just be a `real_t *` by the time it gets here)

inst/include/dust/dust.hpp
richfitz: I'm voting to hold off on this for now - we can add it if we need it

inst/include/dust/dust.hpp
richfitz: I think we can change to add auxillary rng state if needed later; it'd be a small change. One possibility is that we add one additional rng here for extra work and never copy that onto the device, then work out how to get that working in serial with the openmp bits

inst/include/dust/interface.hpp
richfitz: certainly no negative. I will check that we do something sensible with no positive weights though. when we hand over to the C++ for this there will be no checking or error handling....

tests/testthat/test-example.R
richfitz: So, all-zero now just doesn't crash as we will cope with that in the filter itself (the situation where that happens the average ll comes out as zero and we stop). I've added a check for at least non-crashing behaviour now

inst/include/dust/dust.hpp
richfitz: Added an explanatory comment now

johnlees (approved):

PR #157, opened 2021-02-10
Implement basic particle filter in dust

### LINKED ISSUE #155: Implement filter on dust object
With a compiled compare and resampling support we can implement the entire particle filter on the dust object. This may be faster (or may not make much of a difference) but will be useful for the GPU work as we will really need to limit how many roundtrips to R we make.

The biggest pain will be things like the history saving.

### LINKED ISSUE #156: Merge sir and sir2 examples
sir2 needs to move into the package and this split is silly now

richfitz:
Very basic particle filter implementation. This is close to a MVP; the history saving was faffy and could have waited but was useful for confirming that things work as expected.

Things not done but which I think would be better as follow-on PRs:

* save restart information (we're not really using that in production yet so not a great loss)
* partial running support (needed for SMC2)
* easier to use interface with automatic reset (inclined to see what shakes out from mcstate)
* handling of impossible likelihoods, particularly with multiple parameters at once

Fixes #155 
Fixes #156  (I was going to this later but it was a bit annoying)


inst/include/dust/dust.hpp
```diff
@@ -45,8 +45,93 @@ struct pars_t {
...
+  size_t size() const {
+    return history_value.size();
+  }
+
+  void advance() {
+    offset_++;
+  }
+
+  std::vector<real_t> history_value;
+  std::vector<size_t> history_order;
```
johnlees, line 125: Should these be private?

inst/include/dust/dust.hpp
```diff
@@ -45,8 +45,93 @@ struct pars_t {
...
+  filter_state(size_t n_state, size_t n_particles, size_t n_data) :
+    n_state_(n_state), n_particles_(n_particles), n_data_(n_data), offset_(0) {
+    resize(n_state, n_particles, n_data);
+  }
+
+  // default constructable
+  filter_state() : filter_state(0, 0, 0) {
+  }
+
+  void resize(size_t n_state, size_t n_particles, size_t n_data) {
```
johnlees, line 70: This doesn't reset, so presume it's if you want to increase the size. Can the size shrink here too?

inst/include/dust/dust.hpp
```diff
@@ -45,8 +45,93 @@ struct pars_t {
...
+    history(ret.begin());
+    return ret;
+  }
+
+  template <typename Iterator>
+  void history(Iterator ret) const {
+    std::vector<size_t> index_particle(n_particles_);
+    for (size_t i = 0; i < n_particles_; ++i) {
+      index_particle[i] = i;
+    }
```
johnlees, line 101: I still like `std::iota`, but this is probably more readable anyway

inst/include/dust/dust.hpp
```diff
@@ -299,14 +388,20 @@ class Dust {
...
 
   std::vector<size_t> resample(const std::vector<real_t>& weights) {
-    std::vector<size_t> idx(n_particles());
+    std::vector<size_t> index(n_particles());
+    resample(weights, index);
+    return index;
+  }
+
+  void resample(const std::vector<real_t>& weights,
+                std::vector<size_t>& index) {
```
johnlees, line 397: Two thoughts:
- Why have the two separate functions here? Is there a case where we call the second with a non-empty index? What does that do?
- Why pass index by reference and require it to be constructed first? Would it not be simpler for this function to make index and return it instead of void?

inst/include/dust/dust.hpp
```diff
@@ -341,6 +435,10 @@ class Dust {
     return _n_pars;
   }
 
+  size_t n_pars_effective() const {
```
johnlees, line 438: `n_pars_size` might be a better name?

inst/include/dust/dust.hpp
johnlees: Looks like the class is initialised with the default constructor, and then resize used to grow this if data is provided

johnlees: So do we always create the class, and then set the data on it? Is there an argument for setting the data when the class is initialised to simplify these constructors? Or is that flexibility needed?

inst/include/dust/dust.hpp
```diff
@@ -363,39 +461,101 @@ class Dust {
...
+      // the case where there is no data and so we do not resample)
+      //
+      // TODO: we should cope better with the case where one filter
+      // has become impossible but others continue, but that's hard!
+      auto wi = weights.begin();
+      for (size_t i = 0; i < n_pars_effective(); ++i) {
+        log_likelihood_step[i] =
+          scale_log_weights<real_t>(wi, n_particles_each);
+        log_likelihood[i] += log_likelihood_step[i];
+        wi += n_particles_each;
```
johnlees, line 521: You could put this increment in the for loop iterator? Or indeed all of the wi initialisation and iteration

johnlees: Or maybe I've just been doing a lot of python recently

inst/include/dust/dust.hpp
```diff
@@ -363,39 +461,101 @@ class Dust {
...
+      }
+
+      // We could move this below if wanted but we'd have to rewrite
+      // the re-sort algorithm.
+      if (save_history) {
+        state(filter_state_.history_value_iterator());
```
johnlees, line 527: The history value vs history order meaning and use are a little unclear to me – could you add a few comments to them?

inst/include/dust/dust.hpp
```diff
@@ -363,39 +461,101 @@ class Dust {
...
+    std::vector<real_t> weights(n_particles);
+    std::vector<size_t> kappa(n_particles);
+
+    if (save_history) {
+      filter_state_.resize(_index.size(), _particles.size(), _data.size());
+      state(filter_state_.history_value_iterator());
+      filter_state_.advance();
+    }
+
+    for (auto & d : _data) {
```
johnlees, line 506: Ok, this iteration motivates the change to a map. The key will be positive integers. Remind me, does it always go up in increments of 1; does it always start from 0? Or does the usage here totally relax this?

An alternative would be two std::vectors, one of data_step, one of std::vector<data_t>. This is what I think we should/will have to use in the CUDA (complications about user defined data_t struct aside). Maybe we should switch to that here, it would change the use of find in `compare_data` above a bit, I think it ends up being very similar to this solution

inst/include/dust/interface.hpp
```diff
@@ -506,19 +506,59 @@ cpp11::sexp dust_compare_data(SEXP ptr) {
...
 
+template <typename T, typename std::enable_if<!std::is_same<dust::no_data, typename T::data_t>::value, int>::type = 0>
+cpp11::sexp dust_filter(SEXP ptr, bool save_history) {
+  using namespace cpp11::literals;
+  Dust<T> *obj = cpp11::as_cpp<cpp11::external_pointer<Dust<T>>>(ptr).get();
+  cpp11::writable::doubles log_likelihood(obj->filter(save_history));
+  cpp11::sexp history;
+  if (save_history) {
+    cpp11::writable::doubles history_data(obj->filter_history().size());
+    obj->filter_history().history(REAL(history_data));
```
johnlees, line 517: Will this work with `real_t = float`?

inst/include/dust/tools.hpp
```diff
@@ -24,4 +24,24 @@ void resample_weight(typename std::vector<real_t>::const_iterator w,
...
+      max_w = std::max(max_w, *wi);
+    }
+  }
+  real_t tot = 0.0;
+  wi = w;
+  for (size_t i = 0; i < n; ++i, ++wi) {
+    *wi = std::exp(*wi - max_w);
+    tot += *wi;
+  }
+  return std::log(tot / n) + max_w;
```
johnlees, line 44: Useful to note that the average is returned, the scaled weights are stored in the passed w

tests/testthat/test-filter.R
```diff
@@ -0,0 +1,95 @@
+context("filter")
+
+test_that("Can run the filter", {
```
johnlees, line 3: Add a test vs the Kalman filter with the volatility model?

johnlees (changes requested):

inst/examples/sir.cpp
```diff
@@ -1,8 +1,11 @@
 class sir {
 public:
   typedef double real_t;
-  typedef dust::no_data data_t;
+  struct data_t {
+    real_t incidence;
+  };
```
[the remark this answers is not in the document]
richfitz: yes. Note that this file was basically sir2.cpp before

inst/include/dust/dust.hpp
richfitz: yes, them being public was a left-over from a tedious bit of bookkeeping checking

inst/include/dust/dust.hpp
richfitz: There's an argument for that, but we'd probably still retain this at the low level to keep the contructors simple and do that initialisation in R

richfitz: > This doesn't reset, so presume it's if you want to increase the size. Can the size shrink here too?

Size can shrink yes, though the memory won't be reclaimed or cause a copy when that happens

> Looks like the class is initialised with the default constructor, and then resize used to grow this if data is provided

exactly

inst/include/dust/dust.hpp
richfitz: Complicated answer here:

* the second version (existing already, pass by reference) is the one that would be used in the inner loop of the particle filter; for that we create the index once and just use it over and over here. We could rewrite this to take an iterator though we've typically reserved that for cases where we do partial writes
* the first version then just wraps that but is used in the R-exposed resample method

inst/include/dust/dust.hpp
richfitz: Not sure. It's the number of "logical" parameter sets; in the case of a non-multi-pars object `n_pars_` is 0 but there are cases where want that number. There's no great name I suspect

inst/include/dust/dust.hpp
richfitz: I wondered about that, but not sure it would be much clearer; the bookkeeping is ick enough as it is

richfitz:
```
      for (size_t i = 0; i < n_pars_effective(); ++i, wi += n_particles_each) {
        log_likelihood_step[i] =
          scale_log_weights<real_t>(wi, n_particles_each);
        log_likelihood[i] += log_likelihood_step[i];
      }
```

I am not convinced :)

inst/include/dust/dust.hpp
richfitz: goes up in arbitrary increments, but always increasing. Can we hold off on a change like that until this bit starts getting moved to the device as it would not be clearer or faster at the moment

inst/include/dust/interface.hpp
richfitz: yes; `REAL(history_data)` is `double *` (that's needed on the R side) and that will go into the `copy_n` with arguments `std::vector<float>::const_iterator`, `size_t`, `double*` which the compiler will handle sensibly

tests/testthat/test-filter.R
richfitz: Done in be13de8; this was actually pretty pleasant to write

inst/include/dust/dust.hpp
```diff
@@ -93,6 +84,24 @@ class filter_state {
...
+  // So to output sensible history we start with a particle and we
+  // look to see where it "came from" in the previous step
+  // (history_index) and propagate this backward in time to
+  // reconstruct what is in effect a multifurcating tree.
+  //
+  // It's possible we could do this more efficiently for some subset
+  // of particles too (give me the history of just one particle) by
+  // breaking the function before the loop over 'k'.
+  //
+  // Note that we tweat history_order and history_value as read-only
```
johnlees, line 103:
```suggestion
  // Note that we treat history_order and history_value as read-only
```

inst/include/dust/dust.hpp
```diff
@@ -93,6 +84,24 @@ class filter_state {
...
+  // translation of the the code). As we proceed we store the values
+  // of particles *before* resampling and then we store the index used
+  // in resampling. We do not resample all the history at each
+  // resample as that is prohibitively expensive.
+  //
+  // So to output sensible history we start with a particle and we
+  // look to see where it "came from" in the previous step
+  // (history_index) and propagate this backward in time to
+  // reconstruct what is in effect a multifurcating tree.
+  //
```
johnlees, line 98:
```suggestion
  // reconstruct what is in effect a multifurcating tree.
  // This is analogous to the particle ancestor concept in the
  // particle filter literature.
  //
```

johnlees (approved):

PR #158, opened 2021-02-12
Non integer dnbinom

richfitz:
Allows size to be noninteger (see tests)


edknock (approved):
Tested, everything working now

PR #159, opened 2021-02-13
Fix set pars

### LINKED ISSUE #125: Improve set_pars
Currently uses a duplicated codepath from initialise and I think leaves the system inconsistent on failure

richfitz:
This PR fixes handling of set_pars

* works for multiple data
* never leaves the model in an inconsistent state (either it will set all parameters or none)
* now requires that the model never changes size (number of parameter sets, size of the state) which we already assume in mcstate

Fixes #125


DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: dust
 Title: Iterate Multiple Realisations of Stochastic Models
-Version: 0.7.6
+Version: 0.7.7
```
johnlees:
```suggestion
Version: 0.7.8
```

inst/include/dust/dust.hpp
```diff
@@ -567,41 +556,63 @@ class Dust {
...
+        p.size();
+      throw std::invalid_argument(msg.str());
+    }
+    if (_particles.size() == n_particles) {
+#ifdef _OPENMP
+      #pragma omp parallel for schedule(static) num_threads(_n_threads)
+#endif
+      for (size_t i = 0; i < n_particles; ++i) {
+        _particles[i].set_pars(p, set_state);
+      }
```
johnlees, line 575: This branch is to save on allocations if np isn't changing?

inst/include/dust/interface.hpp
```diff
@@ -41,6 +41,8 @@ cpp11::sexp create_matrix(size_t nrow, size_t ncol, T& pars);
 template <typename T>
 cpp11::sexp create_array(const std::vector<size_t>& dim, T& pars);
 
+inline void check_pars_multi(cpp11::list pars, size_t n);
```
johnlees, line 44: inline 👀 
(I approve)

inst/include/dust/dust.hpp
```diff
@@ -567,41 +556,63 @@ class Dust {
...
+#endif
+      for (size_t i = 0; i < _n_particles_total; ++i) {
+        _particles[i].set_pars(p[i / n_particles], set_state);
+      }
+    } else {
+      _particles.clear();
+      _particles.reserve(n_particles * _n_pars);
+      for (size_t i = 0; i < _n_pars; ++i) {
+        for (size_t j = 0; j < n_particles; ++j) {
+          _particles.push_back(p[i]);
```
johnlees, line 612: Could also have openmp here if you s/reserve/resize/ and insert at `i * n_particles + j` (but I'm easy, not sure how much difference it would make anyway)

johnlees (approved):

inst/include/dust/dust.hpp
richfitz: yes, it also lets us do this in parallel (the previous branch can't be done in parallel because we use push_back and we can't resize first because the particles aren't default-constructible)

inst/include/dust/interface.hpp
richfitz: has to be inline to avoid multiple definition errors :)

inst/include/dust/dust.hpp
richfitz: I looked into this and it's hard if we can't default construct. there are some workarounds to make push_back work in parallel but they are all grim

DESCRIPTION
richfitz: You managed to catch this just before I fixed that (we had to quickly sneak a fix in for a broken density function)

inst/include/dust/dust.hpp
johnlees: I previously tried one work-around using `#pragma critical` to merge lists made by each thread. Was no faster, and sometimes segfaulted in other environments

johnlees: (so I removed it)

richfitz: This makes me pleased I did not try! This never shows up as a big timesink normally

PR #160, opened 2021-02-13
Add simulate method, replacing dust_simulate

### LINKED ISSUE #100: Add iterate method to dust object?
Something like `$dust_iterate()` but as a method?

### LINKED ISSUE #119: dust_simulate should optionally return all state
enables restart

### LINKED ISSUE #121: dust_simulate should verify state length
Currently it will accept invalid length state leading to crashes

richfitz:
This PR adds a simulate method to the object, removing the placeholder `dust_simulate` and `dust_iterate` functions (they're deprecated for now but retained so that existing work can continue).

~Builds on #159, so that must be merged first~
Fixes #100 
Fixes #119 (obsoleted)
Fixes #121 (obsoleted)


inst/include/dust/dust.hpp
```diff
@@ -324,6 +324,22 @@ class Dust {
     }
   }
 
+  std::vector<real_t> simulate(const std::vector<size_t>& step_end) {
```
johnlees, line 327: So this now using multi_pars, rather than `std::vector<dust::pars_t<T>>& pars`? Seems like a good idea

johnlees: What's the difference now between `simulate` and `run` on the dust object? Perhaps a comment could be added?

johnlees (approved):

PR #164, opened 2021-02-15
Start on interleave support

### LINKED ISSUE #115: Optional support for running interleaved models
Ideally we will fall back on sensible defaults, but we need to be able support interleaved models, such as those that are targeting the GPU. @johnlees has most of the prototype of this working, and https://github.com/mrc-ide/odin.dust/issues/37 contains the upstream support in odin.dust.

Probably best is to start with a hand-coded dust model taken from the prototype and get the support working, then move to get odin.dust working

richfitz:
This PR lays the groundwork for bringing gpu support back into the package. It is built off of the code in https://github.com/mrc-ide/dustgpu/tree/f1d365233024970fb7adcfe8a5168f4bd6ad2709 but does not actually bring in the GPU support yet.

The hope here is that this will give enough for me to target odin.dust at, while giving us something that will act as a bridge to pull in more of the GPU support.

I'm opting to start on the run kernel as it's the most interesting. The current version differs significantly from our previous attempt at this due to the effect of the shared memory update we did in dust 0.7.0. This actually significantly simplifies the interface as we no longer need to copy the mutable internal state between host and device (we'll document this contract somewhere). It adds a slight complication for multi-parameter dust models, but I think that will sort itself out by the time we get those copied onto the device anyway.

I've taken a few liberties with naming, moving things to snake_case. There's room to update further but it'll get hard if we deviate too far from dustgpu.

I added support to the "variable" model because it's really easy to work with and we don't reference it in the examples yet. I do not want to complicate the walk or sir models much yet.

TODO:

- [x] in run_device we have two "zeros" for the size of scratch integer and real storage. The model needs to tell us what these should be, but none of our models use it yet. I might do that in conjunction with the odin.dust PR to add support for generating this sort of model

NOTE: codefactor is complaining because the dust.hpp and interface.hpp files have become messy. I agree with it but need to do that in a different PR or this will be impossible to follow

Fixes #115


inst/examples/variable.cpp
```diff
@@ -38,6 +38,25 @@ class variable {
...
+// automatically based on the existance of the template below though.
+template <>
+struct dust::has_gpu_support<variable> : std::true_type {};
+
+template <>
+void update_device<variable>(size_t step,
+                             const dust::interleaved<variable::real_t> state,
+                             dust::interleaved<int> internal_int,
+                             dust::interleaved<variable::real_t> internal_real,
+                             dust::shared_ptr<variable> shared,
```
johnlees, line 51: I would suggest that this signature becomes:
```suggestion
                             real_t * shared_real,
                             int * shared_int,
```
But not sure this is possible in this PR

inst/examples/variable.cpp
```diff
@@ -38,6 +38,25 @@ class variable {
...
+                             dust::rng_state_t<variable::real_t>& rng_state,
+                             dust::interleaved<variable::real_t> state_next) {
+  for (size_t i = 0; i < shared->len; ++i) {
+    state_next[i] =
+      dust::distr::rnorm(rng_state, state[i] + shared->mean, shared->sd);
```
johnlees, line 56: These shared members access would then work the same way as the unpacking of internal state from before

inst/examples/variable.cpp
johnlees: The generation of these would be in `run_particles`

inst/include/dust/containers.hpp
```diff
@@ -0,0 +1,149 @@
...
+
+  // Constructor to allocate empty memory
+  device_array(const size_t size) : size_(size) {
+    data_ = (T*) std::malloc(size_ * sizeof(T));
+    if (!data_) {
+      // This is not tested (or easily testable without mocking) but
+      // simple enough. This error will be caught by cpp11
+      //
+      // TODO: we might use `new` here which will throw automatically?
+      throw std::bad_alloc(); // # nocov
```
johnlees, line 29: I'm happy if you want to just change these to `new` and `delete`  now – the use of malloc was for direct CUDA conversion, but we know what to do with that so it's less important than it was in the first attempt

inst/include/dust/dust.hpp
```diff
@@ -15,12 +15,52 @@
...
 #endif
 
+
+// TODO: move these into a utilities file
+// TODO: make safe to use with references
+//
+// NB: these functions expect pointer for dest and ref for src,
+// so if applying to a vector use .data().
+// If the vector is passed in this will still compile due to the template,
+// but as not passed by ref will update a copy not the vector itself
```
johnlees, line 25: Don't know if there's any template magic that would give a compiler error in this case?
(this was one difficulty I mentioned)

inst/include/dust/dust.hpp
```diff
@@ -141,8 +181,66 @@ class filter_state {
...
+
+// We need to compute the size of space required for integers and
+// reals on the device, per particle. Because we work on the
+// requirement that every particle has the same dimension we pass an
+// arbitrary set of shared parameters (really the first) to
+// device_work_size. The underlying model can overload this template
+// for either real or int types and return the length of data
+// required.
+template <typename T>
+size_t device_work_size_int(typename dust::shared_ptr<T> shared) {
```
johnlees, line 205: Can we clarify the name here that this is device_shared_size? Not the internal size

inst/include/dust/dust.hpp
```diff
@@ -141,8 +181,66 @@ class filter_state {
...
+                   dust::interleaved<typename T::real_t> internal_real,
+                   dust::shared_ptr<T> shared,
+                   dust::rng_state_t<typename T::real_t>& rng_state,
+                   dust::interleaved<typename T::real_t> state_next);
+
+template <typename T>
+void run_particles(size_t step_from, size_t step_to, size_t n_particles,
+                   size_t n_state, size_t n_int, size_t n_real, size_t n_pars,
+                   typename T::real_t * state,
+                   std::vector<dust::shared_ptr<T>> shared,
```
johnlees, line 241: This will need to be something simple (number/pointer) as this is the global/kernel signature. I suspect we may have:
```
real_t * shared_real
int * shared_int
int * shared_index
```
Shared_index is probably just the pars index vector already in the dust object

inst/include/dust/dust.hpp
```diff
@@ -459,11 +589,18 @@ class Dust {
...
+    refresh_host();
     return _rng.export_state();
   }
 
   void set_rng_state(const std::vector<uint64_t>& rng_state) {
+    // TODO: In dustgpu we don't do a refresh_host() here but I think
+    // we need to. Otherwise if the canonical state is on the device
+    // it's out of sync. However, this is an issue for other things
+    // (e.g., set_state()) so we might need to think about this more
+    // generally.
```
johnlees, line 601: Yeah I think that was an error – happy with this

inst/include/dust/dust.hpp
```diff
@@ -639,6 +808,201 @@ class Dust {
...
+                   std::vector<dust::shared_ptr<T>> shared,
+                   uint64_t * rng_state) {
+  typedef typename T::real_t real_t;
+  const size_t n_particles_each = n_particles / n_pars;
+
+  // Allocation for our internal storage. We won't ever move this to
+  // or from the device so we just pass the size.
+  std::vector<real_t> state_next_v(n_particles * n_state);
+  std::vector<int> internal_int_v(n_particles * n_int);
+  std::vector<real_t> internal_real_v(n_particles * n_real);
```
johnlees, line 905: This still needs to be allocated from the host code (even if we are no longer using memset/memcpy), so I think these lines should go back into the dust object, become device_arrays, which are created in initialise. They then need to be passed as pointers to this function. It may make sense to have another struct, like you've done to pack the state + rng?

You can't call CUDA API functions from device or global code sadly.

inst/include/dust/dust.hpp
```diff
@@ -639,6 +808,201 @@ class Dust {
...
+  // omp here
+  for (size_t i = 0; i < n_particles; ++i) {
+    dust::interleaved<real_t> p_state(state, i, n_particles);
+    dust::interleaved<real_t> p_state_next(state_next, i, n_particles);
+    dust::interleaved<int> p_internal_int(internal_int, i, n_particles);
+    dust::interleaved<real_t> p_internal_real(internal_real, i, n_particles);
+    dust::interleaved<uint64_t> p_rng(rng_state, i, n_particles);
+    // TODO: this needs work before moving to the device, but it might
+    // not be that bad in practice.
+    dust::shared_ptr<T> p_shared = shared[i / n_particles_each];
```
johnlees, line 920: This is where we'll need some extra code to deal with the blocks (before the for loop)

inst/include/dust/dust.hpp
johnlees: Actually I checked this and `cudaMalloc` and `cudaFree` are both available from device code too, but we still don't want to call these at every kernel entry, so I still think they should be moved back to the dust object.

inst/include/dust/dust.hpp
```diff
@@ -639,6 +808,201 @@ class Dust {
...
+    // original place, but an on odd number of steps the passed
+    // variables need to be swapped.
+  }
+
+  // This is going to require a bit of work to get exactly right as
+  // sometimes we hold the correct data
+  if ((step_end - step_start) % 2 == 1) {
+    // this is easily done without stl with memcpy:
+    // memcpy(state, state_next, sizeof(real_t) * n_state);
+    std::copy_n(state_next, n_state * n_particles, state);
```
johnlees, line 951: I didn't think that you could do either of these from a kernel, which reminds me why I did the pointer swap here. 

Checking, there is `cudaMemcpyAsync` and `cudaMemsetAsync` which can be called in a kernel. However we might want to be careful with this – is calling them from one thread efficient, or would it be better to call these from the host?

inst/include/dust/dust.hpp
```diff
@@ -639,6 +808,201 @@ class Dust {
...
+                       p_state,
+                       p_internal_int,
+                       p_internal_real,
+                       p_shared,
+                       rng_block,
+                       p_state_next);
+      std::swap(p_state, p_state_next);
+      // dust::interleaved<real_t> tmp = p_state;
+      // p_state = p_state_next;
+      // p_state_next = tmp;
```
johnlees, line 934: There is a swap in thrust, but not CUDA, so we probably want to revert this change

inst/include/dust/rng.hpp
```diff
@@ -80,6 +81,30 @@ class pRNG { // # nocov
...
+}
+
 }
 
 #endif
+
+template <typename T, typename U = typename T::real_t>
+U unif_rand(T& state) {
+  const uint64_t value = xoshiro_next(state);
+  return U(value) / U(std::numeric_limits<uint64_t>::max());
```
johnlees, line 109: I liked your idea about the numeric limits helper class. Shall we note this down somewhere?

johnlees (changes requested):
Three main points:
- The shared_t/shared_ptr won't work and have added some thoughts on alternatives here. Happy to leave this for a subsequent PR however, as this works for the CPU.
- I think the yi/internal_real/internal_int initialisation need to go back into the dust class
- We need to be careful about doing memcpy from a kernel. It may be fine, but we should check.

DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: dust
 Title: Iterate Multiple Realisations of Stochastic Models
-Version: 0.7.9
+Version: 0.7.10
```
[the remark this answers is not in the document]
richfitz: yeah, I was thinking about 0.8 as soon as something *can* be run on the GPU, then probably 0.9 once sircovid can be

inst/examples/variable.cpp
richfitz: yeah, that's not possible yet. Take a look [here](https://github.com/mrc-ide/sircovid/blob/master/inst/dust/basic.cpp#L72-L311) at the shared memory in the small sircovid basic model; we've got a motley collection of ints, reals and vectors of reals

inst/examples/variable.cpp
richfitz: Given the mess above, it would be nice if we could use a pointer to a struct even if that struct contains just int, real_t, int* and real_t* (no vectors) with the allocation handled elsewhere. Otherwise the unpack at the beginning of the function is going to be quite painful

inst/include/dust/containers.hpp
richfitz: I'll do this now I think

inst/include/dust/dust.hpp
richfitz: yes, you can fix that with a static assert I think provided we can find a trait that suits the type. Will think about this

inst/include/dust/dust.hpp
richfitz: This is the internal size

inst/include/dust/rng.hpp
richfitz: Added here: https://github.com/mrc-ide/dust/issues/165

inst/include/dust/dust.hpp
richfitz: we could implement our own :)

inst/include/dust/dust.hpp
richfitz: this one you might have to explain to me as I am a bit in the dark about how these memory allocations work in the CUDA model

R/dust_class.R
```diff
@@ -86,7 +86,13 @@ dust_class <- R6::R6Class(
...
     ##' @param step_end Step to run to (if less than or equal to the current
     ##' step(), silently nothing will happen)
-    run = function(step_end) {
+    ##'
+    ##' @param device **Experimental!**: This argument may allow running on
+    ##' a GPU once support is finished, if the model supports it, and if
+    ##' the model is compiled appropriately (and assuming you have a
+    ##' suitable GPU). At present it exists for testing and will run
+    ##' slower than running with `device = TRUE`.
+    run = function(step_end, device = FALSE) {
```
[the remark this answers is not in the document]
richfitz: done in ffb0915

inst/examples/variable.cpp
richfitz: As discussed offline (well different online) we'll mock up some ideas around this starting with a single parameter set and see how we get on

inst/include/dust/containers.hpp
richfitz: Done in 4605ef6

inst/include/dust/dust.hpp
richfitz: Done in 9f2bfd1, you will probably like this

inst/include/dust/dust.hpp
richfitz: Fixed in 0e367ea

inst/include/dust/dust.hpp
richfitz: Fixed in ad21b25

inst/include/dust/dust.hpp
richfitz: Fixed in ad21b25

inst/include/dust/dust.hpp
richfitz: Leaving this as is for now as we can change that around easily later

inst/include/dust/dust.hpp
```diff
@@ -316,12 +420,59 @@ class Dust {
...
+    // interleaved variables updated, but the interleaved variables
+    // passed in have not yet been updated.  If an even number of
+    // steps have been run state will have been swapped back into the
+    // original place, but an on odd number of steps the passed
+    // variables need to be swapped.
+    if ((step_end - step_start) % 2 == 1) {
+      // this is easily done without stl with memcpy:
+      // memcpy(_device_data.y.data(), y_next.data(), sizeof(real_t) * n_state);
+      real_t *dst = _device_data.y.data();
+      std::copy_n(y_next.data(), n_state * n_particles, dst);
```
johnlees, line 471: Still a long copy in the kernel – can we add a note to be careful when porting this to CUDA?
Some discussion of this here: https://stackoverflow.com/a/49037139

We should also look out for this in the profile: how long would a copy of 100,000 particles with 1,000 floats of state take? Could well be longer than the update.
Alternatively just leaving the copy as async may help hide any latency here

johnlees (changes requested):

inst/include/dust/containers.hpp
```diff
@@ -3,8 +3,8 @@
 
 #include <cstdint>
 #include <cstddef>
-#include <cstdlib> // malloc
 #include <cstring> // memcpy
+#include <new>
```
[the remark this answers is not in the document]
richfitz: almost certainly dragged in by something else :)

inst/include/dust/dust.hpp
```diff
@@ -892,22 +914,13 @@ class Dust {
 template <typename T>
 void run_particles(size_t step_start, size_t step_end, size_t n_particles,
                    size_t n_state, size_t n_int, size_t n_real, size_t n_pars,
-                   typename T::real_t * state,
+                   typename T::real_t * state, typename T::real_t * state_next,
+                   int * internal_int, typename T::real_t * internal_real,
```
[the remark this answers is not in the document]
richfitz: that makes me wonder if the prototype is needed

inst/include/dust/dust.hpp
richfitz: this is now a swap

inst/include/dust/dust.hpp
```diff
@@ -442,9 +443,34 @@ class Dust {
...
-                     n_state_full(), n_int, n_real, n_pars_effective(),
-                     _device_data.y.data(), _shared, _device_data.rng.data());
+
+    // Allocation for our internal storage. We won't ever move this
+    // back to the cpu
+    const size_t n_particles = _particles.size();
+    const size_t n_state = n_state_full();
+    std::vector<real_t> y_next(n_particles * n_state);
+    std::vector<int> internal_int(n_particles * n_int);
+    std::vector<real_t> internal_real(n_particles * n_real);
```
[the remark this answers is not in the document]
richfitz: Moved this into the device_state object now

johnlees (approved):

PR #168, opened 2021-02-18
Proper(ish) handling of shared parameters for gpu models

richfitz:
I toyed with the struct approach, and I think it could work, but it's fiddly to get right. We can start like this at least.

See variable.cpp for the new interface - another couple of template functions that the user needs to define, including one for translating the nice shared type into the plain shared type. With this we do one big allocation, and we can certainly write it to do one copy to the device.

Fixes #116


inst/examples/variable.cpp
```diff
@@ -43,19 +43,44 @@ class variable {
...
+}
+
+template <>
+void device_shared_copy<variable>(dust::shared_ptr<variable> shared,
+                                  int * shared_int,
+                                  variable::real_t * shared_real) {
+  typedef variable::real_t real_t;
+  shared_int = dust::shared_copy<int>(shared_int, shared->len);
+  shared_real = dust::shared_copy<real_t>(shared_real, shared->mean);
+  shared_real = dust::shared_copy<real_t>(shared_real, shared->sd);
```
johnlees, line 64: What would be the method to write this as a single H->D copy if we needed to do that? Could we 'stage' a std::vector on the host here, copy into it in parts like this, then copy the whole thing over to the device at the end?

inst/include/dust/dust.hpp
```diff
@@ -810,16 +858,44 @@ class Dust {
...
+  typename std::enable_if<dust::has_gpu_support<U>::value, void>::type
+  update_device_shared() {
+    const size_t n_shared_int = _device_data.n_shared_int;
+    const size_t n_shared_real = _device_data.n_shared_real;
+    for (size_t i = 0; i < _shared.size(); ++i) {
+      // TODO: John, I've not used set_array here at least for now;
+      // this should work fine for the CPU but might need a bit of
+      // support on the GPU? If needed we can stage the copy on the
+      // cpu by putting it into one std::vector and then doing two big
+      // set_array calls. We never need to worry about the return leg.
```
johnlees, line 894: great minds (cf above comment)

johnlees: I think it probably depends on how often it gets called, we might get away with it if it's just done at the start. But thinking about the carehomes model (did you say about 1000 items in here?) maybe that will become a problem, especially if we start writing this more frequently, which seems possible with some inference methods

inst/include/dust/dust.hpp
```diff
@@ -931,15 +1009,18 @@ void run_particles(size_t step_start, size_t step_end, size_t n_particles,
     // TODO: this needs work before moving to the device, but it might
     // not be that bad in practice. We'll need some extra code to deal
     // with the blocks (before the loop) too.
-    dust::shared_ptr<T> p_shared = shared[i / n_particles_each];
+    const int j = i / n_particles_each;
+    const int * p_shared_int = shared_int + j * n_shared_int;
+    const real_t * p_shared_real = shared_real + j * n_shared_real;
```
johnlees, line 1014: This is nice and should easily be changed to the CUDA pattern I was thinking of

johnlees: I think it will work as-is in fact, but we can speed it up by loading into L1 shared if it will fit

inst/include/dust/dust.hpp
johnlees: An argument for doing this now is that we found profiling very difficult when we had loads of CUDA API calls

inst/include/dust/dust.hpp
johnlees: And makes it easy to make conditional on n_particles_each too

johnlees (approved):

inst/examples/variable.cpp
richfitz: yes, exactly. That can be done in parallel too if we need to

inst/include/dust/dust.hpp
richfitz: OK, I've added a staged copy here, which should make the cuda work easier (1dc743e)

johnlees (approved):

PR #181, opened 2021-02-25
New vignette about general design principles

### LINKED ISSUE #161: Add design vignette
This should cover

* why is a package even needed for this sort of thing?
* how does the parallel RNG work?
* how does the gpu interface work?

richfitz:
Very light on code, but contains things that I've told people verbally

Also updated vignette titles and did some spellchecking

Fixes #161


vignettes/design.Rmd
```diff
@@ -0,0 +1,210 @@
...
+* In addition to the above, run for many parameter sets at once possibly with one particle per parameter, possibly with many per parameter
+* In addition to the above, the parameters themselves are grouped into blocks
+
+There book-keeping for this can get tedious and error prone if done by hand. In dust, we try and restrict concern about this to a few points, and for the simulation itself -- the interaction that we expect to take the longest in any interesting model -- we just run a big loop over time and all particles no matter what type of structure they might represent from the above.
+
+See `vignette("multi")` for details of interacting with different ways that you might want to structure your simulations.
+
+## Parallelisation
+
+Once we we're running multiple simulations at once, things will get slower and we might look to parallelism to try and speed up our simulations.
```
RaphaelS1, line 63: Typo - "Once we we're"

vignettes/design.Rmd
```diff
@@ -0,0 +1,210 @@
...
+
+```r
+runif(10)
+```
+
+that would draw the 10 numbers in parallel rather than in series. When drawing a random number there is a "[side effect](https://en.wikipedia.org/wiki/Side_effect_(computer_science))" of updating the random number state. That is because the random number stream is *also* a Markov chain!
+
+As such it makes sense (to us at least) to store the state of each stream's random number generator separately, so if we have `n` particles within a `dust` object we have `n` separate streams, and we might think of the model state as being the state that is declared by the user as a vector of doubles alongside the random number state. During each model step, the model state is updated and so is the random number state.
+
+This might seem wasteful, and if we used the [Mersenne Twister](https://en.wikipedia.org/wiki/Mersenne_Twister) it would be to some degree as each particle would require 2560 bytes of additional state. In contrast the [xoshiro](http://prng.di.unimi.it/) generators that we 256 bytes of state; the same as 4 double-precision floating point numbers. So for any nontrivial simulation it's not a very large overhead.
```
RaphaelS1, line 75: Typo: " generators that we 256 bytes of state"

vignettes/design.Rmd
```diff
@@ -0,0 +1,210 @@
...
+
+Setting the seed for these runs is not trivial, particularly as the number of simultaneous particles increase. If you've used random numbers with the [future](https://cran.r-project.org/package=future) package you may have seen it raise a warning if you do not configure it to use a "L'Ecuyer-CMRG" which adapts R's native random number seeds to be safe in parallel.
+
+The reason for this that if different streams start from seeds that are set via poor heuristics (e.g., system time) they might be exactly the same. If they were set randomly, then they might collide (see [John Cook's description of the birthday paradox here](https://www.johndcook.com/blog/2016/01/29/random-number-generator-seed-mistakes/)) and if they are picked sequentially there's no guarantee that these streams might not be correlated.
```
RaphaelS1, line 79: Typo: "The reason for this that"

RaphaelS1 (commented):
Just flagged some minor typos

vignettes/design.Rmd
```diff
@@ -0,0 +1,210 @@
...
+lang_output <- function(x, lang) {
+  cat(c(sprintf("```%s", lang), x, "```"), sep = "\n")
+}
+cc_output <- function(x) lang_output(x, "cc")
+r_output <- function(x) lang_output(x, "r")
+plain_output <- function(x) lang_output(x, "plain")
+set.seed(1)
+```
+
+Fundamentally, stochastic Monte Carlo models are extremely simple. Consider a random walk in one dimension - we might write that in base R functions by creating a function that takes a current state `state` and a list of parameters:
```
johnlees, line 26:
```suggestion
Fundamentally, using a computer to create realisations from stochastic Monte Carlo models is extremely simple. Consider a random walk in one dimension - we might write that in base R functions by creating a function that takes a current state `state` and a list of parameters:
```

vignettes/design.Rmd
```diff
@@ -0,0 +1,210 @@
...
+
+## Efficient state handling
+
+For nontrivial systems, we often want to record a subset of states - potentially a very small fraction of the total states computed. For example, in our [sircovid](https://mrc-ide.github.io/sircovid/) model we track several thousand states (representing populations in various stages of disease transmission, in different ages, with different vaccination status etc), but most of the time we only need to report on a few tens of these in order to fit to data or to examine key outputs.
+
+Reducing the number of state variables returned at different points in the process has several advantages
+
+* Saving space: if you run a model with 2000 states, and 1000 replicates and record trajectories over 100 timesteps, that represents 100 million double precision numbers, or 1.6 GB of memory or disk (before compression). These are not unrealistic numbers, but would make even a simple sensitivity analysis all-but impossible to work with.
```
johnlees, line 106:
```suggestion
* Saving space: if you run a model with 2000 states, 1000 replicates and record their trajectories over 100 timesteps, that represents 100 million double precision numbers, or 1.6 GB of memory or disk (before compression). These are not unrealistic numbers, but would make even a simple sensitivity analysis all-but impossible to work with.
```

vignettes/design.Rmd
```diff
@@ -0,0 +1,210 @@
...
+* Saving time: copying data around is surprisingly slow
```
johnlees, line 107:
```suggestion
* Saving time: copying data around is surprisingly slow.
```

vignettes/design.Rmd
```diff
@@ -0,0 +1,210 @@
...
+
+To enable this, you can restrict the state returned by most methods; some by default and others when you call them.
+
+* The `$run()` and `$simulate()` methods move the system forwards in time and returns the state at that point; it uses an index set into the object with `$set_index()`. The intention here is that these would be repeatedly called and so we validate the index once and use it over and over.
+* The `$state()` method returns the model state and accepts an argument `index` as the state to return
+
+In both cases, if `index` was named then the returned state carries these names as its rownames.
+
+The ordering of the state is important; we always have dimensions that will contain
```
johnlees, line 116:
```suggestion
The ordering of the state is important; we always have dimensions that will contain:
```

vignettes/design.Rmd
```diff
@@ -0,0 +1,210 @@
...
+
+1. the states within a single particle
```
johnlees, line 118:
```suggestion
1. the model states within a single particle
```

vignettes/design.Rmd
```diff
@@ -0,0 +1,210 @@
...
+2. the particles within a time (may be several dimensions; see `vignette("multi")`)
```
johnlees, line 119:
```suggestion
2. the particles within a time-step (may be several dimensions; see `vignette("multi")`)
```

vignettes/design.Rmd
```diff
@@ -0,0 +1,210 @@
...
+3. the time dimension if using
+
+This is to minimise churn during writing, and to help with concatenation. Multiple particles data is stored consecutively and read and written in order. Each time step is written at once. And you can append states from different times easily. The `aperm()` function will be useful for reshaping this output to a different dimension order if you require one.
```
johnlees, line 122: Define churn?

vignettes/design.Rmd
```diff
@@ -0,0 +1,210 @@
...
+
+## Useful verbs
+
+We try and provide verbs that are useful, given that the model presents a largely opaque pointer to model state. These are driven by our needs for running a particle filter.
+
+Normally we have in the object several things:
+
+* the random number state: this is effectively a matrix of 64-bit integers
+* the model state: this is effectively a matrix of floating point numbers, typically `double`s
+* the model parameters: this is specific to the model in question and these are likely shared across multiple particles). This is presented to all particles within a group as an immutable pointer, which allows safe simultaneous access by multiple threads
```
johnlees, line 134:
```suggestion
* the model parameters: this is specific to the model in question (and these are likely shared across multiple particles). This is presented to all particles within a group as an immutable pointer, which allows safe simultaneous access by multiple threads
```

vignettes/design.Rmd
```diff
@@ -0,0 +1,210 @@
...
+* the model _internal state_: this is also specific to the model and dust has no control over this. Because each particle has its own state they can safely write to it when running parallel (compare with parameters). Typically this space is *allocated* at model construction. For models that come from [odin.dust](https://mrc-ide.github.io/odin.dust/) this is used as "scratch" space that we write to during an update but which does not conceptually persist between steps.
```
johnlees, line 135: Add an example of an internal state?

vignettes/design.Rmd
```diff
@@ -0,0 +1,210 @@
...
+
+There is one subtlety about internal state: we assume that the state entirely specifies a model in a Markov process and we don't guarantee that models with mutable internal state that is not conceptually discarded each iteration. This is because if we reorder particles what we really do is reorder the *state vector* and not this internal state. This prevents implementing things like models with "delays" in the current design. We may relax constraint this if it is needed.
```
johnlees, line 137:
```suggestion
There is one subtlety about internal state: we assume that the state entirely specifies a model in a Markov process, and so we don't guarantee that models with mutable internal state will not be discarded between each iteration. This is because if we reorder particles what we really do is reorder the *state vector* and not this internal state. This prevents implementing things like models with "delays" in the current design. We may relax constraint this if it is needed.
```

vignettes/design.Rmd
```diff
@@ -0,0 +1,210 @@
...
+
+The sorts of verbs that we need given this include:
```
johnlees, line 139:
```suggestion
Given this, the sorts of verbs that we need include:
```

vignettes/design.Rmd
```diff
@@ -0,0 +1,210 @@
...
+
+* Setting model state (`$set_state`) - leaves RNG state and parameters untouched but replaces model state for all particles. This is useful for model initialisation and for performing arbitrary model state changes.
+* Running the model up to a time point (`$run`) - runs the model's `update` method as many times as required to reach the new time point, returning the model state at this time point. This is useful where you might want to change the system at this time point, then continue.
+* Running the model and collecting history (`$simulate`) - as for `$run` but also collects partial state at a number of times along the way. This always has one more dimension than `$run` (being time) and the two functions coexist so that dimensionality is easy to program against.
+* Reset particles and their parameters (`$reset` and `$set_pars`) which
+
+In addition, we have more specific methods oriented towards [particle filtering](https://en.wikipedia.org/wiki/Particle_filter)
```
johnlees, line 146:
```suggestion
In addition, we have more specific methods oriented towards [particle filtering](https://en.wikipedia.org/wiki/Particle_filter):
```

vignettes/design.Rmd
```diff
@@ -0,0 +1,210 @@
...
+
+* Reordering the particles (`$reorder`) - shuffles particle state among particles within a parameter set. This is useful for implementing a resampling algorithms and updates only the state (as for `$set_state`, leaving RNG state and internal state untouched)
+* Resampling particles according to some weight vector (`$resample`) which implements a bootstrap sampling algorithm on top of `$reorder`
+* Run a bootstrap particle filter (`$filter`) which implements a bootstrap particle filter using the above methods, in the case where the model provides a compare function. This is likely to be a bit low level for direct use, and is better approached via the interface in [mcstate](https://mrc-ide.github.io/mcstate/)
```
johnlees, line 150:
```suggestion
* Run a bootstrap particle filter (`$filter`) which implemented using the above methods, in the case where the model provides a compare function. This is likely to be a bit low level for direct use, and is better approached via the interface in [mcstate](https://mrc-ide.github.io/mcstate/)
```

johnlees (changes requested):

vignettes/design.Rmd
richfitz: done

johnlees (approved):

PR #184, opened 2021-02-26
Update vignette issues

### LINKED ISSUE #98: Vignette needs updating

### LINKED ISSUE #182: More vignette issues
in dust.Rmd:

* "There are two main parts here; the class (walk) and an interface function (dust_data)." => dust_pars
* "the type shared_t is whatever data your model needs that is shared across all particles, and unmodified during the model update (most likely parameters), and an object of this type will be passed to your constructor as the .shared field to pars (see dust_pars below)" incorrect for multi-dust
* " It does not itself provide any statistical support such as particle filter; see the mcstate package for that" grammar, and also not correct
* "We use as many random number generators as their are threads, so if you run fewer particles or more, increase the threads or decrease, the results will be the same." => as there are *particles*
* "There is not much different here to the version that is shown before for the walk model;" manifestly untrue
* "To get infomation about the parameters, use coef() on the generator" - not followed by a real code block
* "In order to run the simulation beginning-to-end, we use dust::dust_iterate(); this function is purely a convenience function and the intention is that you would write your own (see mcstate for example)." Deprecated and raises warning....
* The reset the model part should be reworked (#98)

richfitz:
Deals with a number of small issues in the main package vignette

Fixes #182
Fixed #98

### DIFF
--- a/vignettes/dust.Rmd
+++ b/vignettes/dust.Rmd
@@ -24,11 +24,12 @@ plain_output <- function(x) lang_output(x, "plain")
 
 Stochastic models can be used in statistical inference via methods such as [particle filtering](https://en.wikipedia.org/wiki/Particle_filter) but doing so requires that the models can be run over and over again very quickly. While R has excellent support for sampling from distributions, it is not necessarily well suited for this sort of problem because R is single threaded, so we are forced to evaluate realisations of the stochastic process in series (one after another) rather than in parallel.
 
-The `dust` package provides tools to help write stochastic models that can be evaluated in parallel. It does not itself provide any statistical support such as particle filter; see the [mcstate](https://mrc-ide.github.io/mcstate) package for that. Instead, it focuses on providing:
+The `dust` package provides tools to help write stochastic models that can be evaluated in parallel. It does not directly provide statistical methods; see the [mcstate](https://mrc-ide.github.io/mcstate) package for that. Instead, it focuses on providing:
 
 * a fast random number generator that can be evaluated in parallel (see `vignette("rng")` for details)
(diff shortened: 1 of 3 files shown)


vignettes/dust.Rmd
```diff
@@ -143,7 +144,7 @@ model <- walk$new(list(sd = 1), 0, 20, n_threads = 1)
 model$run(100)
 ```
 
-We use as many random number generators as their are threads, so if you run fewer particles or more, increase the threads or decrease, the results will be the same.
+We use as many random number generators as their are particles, so if you run fewer particles or more, increase the threads or decrease, the results will be the same (see `vignette("design")` for more on this).
```
johnlees, line 147:
```suggestion
We use as many random number generators as there are particles, so if you run fewer particles or more, increase the threads or decrease, the results will be the same (see `vignette("design")` for more on this).
```

vignettes/dust.Rmd
```diff
@@ -153,7 +154,7 @@ Consider now an SIR model (Susceptible - Infected - Recovered). This sort of mod
 cc_output(readLines(dust:::dust_file("examples/sir.cpp")))
 ```
 
-There is not much different here to the version that is shown before for the walk model;
+There is a bit different here to the version that is shown before for the walk model, but the core parts are only slightly different:
```
johnlees, line 157:
```suggestion
There a few changes here compared to the version that was shown before for the walk model, but the core parts are only slightly different:
```

johnlees (approved):

PR #194, opened 2021-03-04
Fix stirling tail approximation

richfitz:
Compare numbers with https://github.com/mrc-ide/dust/blob/v0.7.17/inst/include/dust/distr/binomial.hpp#L53-L57

This generates with exactly the formula used and also splits apart the double/float difference from the gpu/cpu split


inst/template/gamma_table.hpp
```diff
@@ -0,0 +1,24 @@
+{{header}}
+#ifndef DUST_DISTR_GAMMA_TABLE_HPP
+#define DUST_DISTR_GAMMA_TABLE_HPP
+
+#ifdef __CUDA_ARCH__
```
johnlees, line 5: Shouldn't this be `__NVCC__`? I'm not totally sure, but as it's not compiling a device function when it reaches these I think CUDA_ARCH will be undef, and so these end up `const` rather than `__constant__` in all cases. Which means they will get stored in global memory instead of the constant cache

scripts/update_gamma_table
```diff
@@ -0,0 +1,44 @@
...
+  writeLines(txt, dest)
+}
+
+glue_whisker <- function(template, data) {
+  glue::glue(template, .envir = data, .open = "{{", .close = "}}",
+             .trim = FALSE)
+}
+
+len_f <- 128L
+len_d <- 10L
```
johnlees, line 18: Was this 16 before, or has it always been 10?

tests/testthat/test-rng.R
```diff
@@ -129,6 +129,20 @@ test_that("binomial random numbers from floats have correct distribution", {
 })
 
 
+test_that("binomial random numbers from floats have correct distribution", {
+  test_rbinom_double <- function(i, m, n, p) {
+    dust_rng$new(1, 1)$rbinom(m, n, p)
+  }
+  m <- 100000
+  n <- 100L
+  p <- 0.1
```
johnlees, line 138: Do one test at np < 10, one at np > 10?

johnlees (changes requested):

inst/template/gamma_table.hpp
richfitz: with __NVCC__ I got warnings saying that we couldn't use `__constant__` from the host

scripts/update_gamma_table
richfitz: always 10

tests/testthat/test-rng.R
richfitz: that exists already - see block above with

```
  m <- 100000
  n <- 958L
  p <- 0.004145
```

johnlees (approved):

PR #199, opened 2021-03-05
Improve device scatter

### LINKED ISSUE #198: Rewrite the scatter kernel
Rather than computing the full scatter index on the host, compute it on the device

johnlees:
Closes #198 

Merge after #200

### DIFF
--- a/inst/include/dust/kernels.hpp
+++ b/inst/include/dust/kernels.hpp
@@ -28,12 +28,13 @@ KERNEL void scatter_device(const size_t* index,
   // offset
   // [3, 1, 3, 2, 3 + 4, 1 + 4, 3 + 4, 2 + 4, 3 + 8, 1 + 8, 3 + 8, 2 + 8]
   // [3, 1, 3, 2, 7, 5, 7, 6, 11, 9, 11, 10]
+  int state_size = n_state * n_particles;
 #ifdef __NVCC__
   // https://developer.nvidia.com/blog/cuda-pro-tip-write-flexible-kernels-grid-stride-loops/
   for (int i = blockIdx.x * blockDim.x + threadIdx.x; i < state_size;
        i += blockDim.x * gridDim.x) {
 #else
-  for (size_t i = 0; i < n_state * n_particles; ++i) {
+  for (int i = 0; i < state_size; ++i) {
 #endif
     const int scatter_index = index[i % n_particles] + (i / n_particles) * n_particles;
     scatter_state[i] = state[scatter_index];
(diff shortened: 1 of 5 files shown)


johnlees:
I think this works. Before:
image


After:
image

So reduces the H->D memcpy at the start, and the time in-between calls. There is obviously still a lot of overhead with this short run, which I assume is caused by the read/copy of floats into R, but we should check in larger models and once the whole particle filter is on the device

johnlees:
Zooming in. Before:
image

After:
image

Kernel now takes 50us rather than 45us, but memcpy goes from 1.42ms to 0.15ms

inst/include/dust/kernels.hpp
```diff
@@ -18,18 +18,29 @@ DEVICE void update_device(size_t step,
...
+                           const size_t n_particles) {
+  // e.g. 4 particles with 3 states ABC stored on device as
+  // [1_A, 2_A, 3_A, 4_A, 1_B, 2_B, 3_B, 4_B, 1_C, 2_C, 3_C, 4_C]
+  // e.g. index [3, 1, 3, 2] with would be
+  // [3_A, 1_A, 3_A, 2_A, 3_B, 1_B, 3_B, 2_B, 3_C, 1_C, 3_C, 2_C]
+  // interleaved, i.e. input repeated n_state_full times, plus a strided
+  // offset
+  // [3, 1, 3, 2, 3 + 4, 1 + 4, 3 + 4, 2 + 4, 3 + 8, 1 + 8, 3 + 8, 2 + 8]
+  // [3, 1, 3, 2, 7, 5, 7, 6, 11, 9, 11, 10]
+  int state_size = n_state * n_particles;
```
richfitz, line 34: out of curiosity what is the motivation for moving to `int` rather than `size_t` here? (with the effect on line 40 with the for loop)

richfitz (commented):
one quick question about the type change - code looks fine though

inst/include/dust/kernels.hpp
johnlees: A micro-optimisation: https://docs.nvidia.com/cuda/cuda-c-best-practices-guide/index.html#loop-counters-signed-vs-unsigned
Though maybe this could have stayed as size_t (and I guess strictly should be const)

richfitz: > Best practices suggest that this optimization be performed after all higher-level optimizations have been completed 

:joy:

richfitz (approved):

PR #200, opened 2021-03-08
Fix use of utils in device code

johnlees:
Fixes two issues:

1. Numeric limits helper functions were not being compiled for use in device code (and compiler error likely suppressed with `__nv_exec_check_disable__`)

For posterity:
Run the kernel, get 'unspecified launch failure'. 
Run in the debugger, end up in CUDBG_ERROR_UNKNOWN_FUNCTION. 
Debugger also craps out because I think because the program counter gets set to 0. [I think the reason is because it is trying to call a branch to a function which has not been compiled (hence PC = 0)]
Specifically, this is epsilon() which is compiled as a host function only.
The fix is to add HOSTDEVICE to all of the numeric limits functions

2. Shared memory copies could be misaligned when `real_t = double` and `n_shared_int % 2 = 1` (address of real_t ptr is a multiple of 4 but not 8).

### DIFF
--- a/inst/include/dust/kernels.hpp
+++ b/inst/include/dust/kernels.hpp
@@ -64,6 +64,8 @@ KERNEL void run_particles(size_t step_start,
   // into __shared__ L1
   extern __shared__ int shared_block[];
   auto block = cooperative_groups::this_thread_block();
+  static_assert(sizeof(real_t) >= sizeof(int),
+                "real_t and int shared memory not alignable");
   if (use_shared_L1) {
     int * shared_block_int = shared_block;
     shared_mem_cpy(block, shared_block_int, p_shared_int, n_shared_int);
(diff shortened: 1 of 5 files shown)


inst/include/dust/kernels.hpp
```diff
@@ -68,7 +71,13 @@ KERNEL void run_particles(size_t step_start,
 
     // Must only have a single __shared__ definition, cast to use different
     // types within it
-    real_t * shared_block_real = (real_t*)&shared_block[n_shared_int];
+    // Furthermore, writing must be aligned to the word length (may be an issue
+    // with int and real, as odd n_shared_int leaves pointer in the middle of an
+    // 8-byte word)
+    assert(sizeof(real_t) >= sizeof(int));
```
richfitz, line 77:
```suggestion
    static_assert(sizeof(real_t) >= sizeof(int),
                  "Swap requires copying");
```

You might need to move this to the top level of the function though if it fails to compile. See https://en.cppreference.com/w/cpp/language/static_assert for details - we use this elsewhere in the package and you just get a compilation failure which is much nicer than a runtime failure

richfitz (changes requested):
What was the issue with rnorm?

johnlees:
> What was the issue with rnorm?

The epsilon utility helper wasn't being compiled on the device, so not an issue with rnorm as such, just that it tried to call this and failed

richfitz (approved):

PR #201, opened 2021-03-12
Save restart (snapshot) information during filter run

### LINKED ISSUE #176: Filter should allow saving restart
Not hard, just a bit more book-keeping

richfitz:
A bit of a terminology change here, but I think this is better and I might update in mcstate too.

Fixes #176


inst/include/dust/dust.hpp
```diff
@@ -520,17 +532,22 @@ class Dust {
       }
 
       // We could move this below if wanted but we'd have to rewrite
-      // the re-sort algorithm.
-      if (save_history) {
-        state(filter_state_.history_value_iterator());
+      // the re-sort algorithm; that would be worth doing I think
```
johnlees, line 535: Can you expand on this/add an issue? Does 'below' mean after running `resample`?

tests/testthat/test-filter.R
```diff
@@ -126,3 +128,63 @@ test_that("scale log weights copes with NaN", {
...
+  dat <- example_filter()
+
+  np <- 10
+  mod <- dat$model$new(list(), 0, np, seed = 10L)
+  mod$set_data(dat$dat_dust)
+
+  ## TODO: handle better doubles here using as_integer
+  step_snapshot <- c(100L, 200L, 400L)
+
+  ## We can perform the entire particle filter manually with the C
```
johnlees, line 145:
```suggestion
  ## We can perform the entire particle filter manually with the C++
```

tests/testthat/test-filter.R
```diff
@@ -126,3 +128,63 @@ test_that("scale log weights copes with NaN", {
   expect_equal(res$weights[[3]], 0)
   expect_equal(res, scale_log_weights(w))
 })
+
+
+## This is badly wrong: we've written into some incorrect memory and
+## we never aligned correctly with the steps! Baby steps.
```
johnlees, line 134: But fixed now? Or where is this bad write happening?

johnlees (changes requested):

inst/include/dust/types.hpp
```diff
@@ -135,6 +133,69 @@ class filter_state {
...
+    n_particles_ = n_particles;
+    n_steps_ = steps.size();
+    offset_ = 0;
+    steps_ = steps;
+    state_.resize(n_state_ * n_particles_ * n_steps_);
+  }
+
+  bool collect(size_t step) {
+    return offset_ < n_steps_ && steps_[offset_] == step;
+  }
```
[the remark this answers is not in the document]
richfitz: It's testing if this is a time that we should collect a shapshot for, I'll add a comment and possibly rename the method to make it clearer

tests/testthat/test-filter.R
richfitz: oops, my tests are always littered with notes to myself like this when refactoring...

inst/include/dust/dust.hpp
richfitz: yeah, exactly. Atm we collect the state, then resort - when reconstructing the tree we rely on that order. But we could do both steps at once which reduces the amount of logic in the function but requires rewriting some very painful code

johnlees (approved):

PR #205, opened 2021-03-15
Tidy up profile start/stop

### LINKED ISSUE #197: Move cuda profiler blocks behind ifdefs
We already have a profile arg to dust_cuda_options, could define an ifdef like `DUST_ENABLE_PROFILE`, then move things (including the Dust destructor!) behind an ifdef

richfitz:
Quite a small quality of life improvement, mostly I like this because it removes the destructor.

Also makes the nvcc options fully configurable - we could theoretically remove the fast math one with this in, but I think that'll be fairly popular

Fixes #197

### DIFF
--- a/R/cuda.R
+++ b/R/cuda.R
@@ -293,7 +293,8 @@ cuda_options <- function(info, debug, profile, fast_math) {
     nvcc_flags <- "-O2"
   }
   if (profile) {
-    nvcc_flags <- paste(nvcc_flags, "-pg --generate-line-info")
+    nvcc_flags <- paste(nvcc_flags, "-pg --generate-line-info",
+                        "-DDUST_ENABLE_CUDA_PROFILER")
   }
   if (fast_math) {
     nvcc_flags <- paste(nvcc_flags, "--use_fast_math")
--- a/R/cuda.R
+++ b/R/cuda.R
@@ -117,6 +117,13 @@ dust_cuda_configuration <- function(path_cuda_lib = NULL,
 ##'   docs](https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html)
 ##'   for more details.
 ##'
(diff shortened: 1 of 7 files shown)


R/cuda.R
```diff
@@ -117,6 +117,13 @@ dust_cuda_configuration <- function(path_cuda_lib = NULL,
 ##'   docs](https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html)
 ##'   for more details.
 ##'
+##' @param flags Optional extra arguments to pass to nvcc. These
+##'   options will not be passed to your normal C++ compiler, nor the
+##'   linker (for that use R's user Makevars system). This can be used
+##'   to do things like tune the maximum number of registers
+##'   (`--maxregcount=x`). If not `NULL`, this must be a character
```
johnlees, line 124:
```suggestion
##'   (`--maxrregcount x`). If not `NULL`, this must be a character
```

(I don't know why...)

inst/include/dust/dust.hpp
```diff
@@ -73,8 +73,10 @@ class Dust {
     }
   }
 
-// This could be removed/commented out eventually
-#ifdef __NVCC__
+  // We only need a destructor when running with cuda profiling; don't
+  // include ond otherwise because we don't actually follow the rule
+  // of 3/5/0
+#ifdef DUST_ENABLE_CUDA_PROFILER
```
johnlees, line 79: Is the `#endif` missing here?

johnlees (changes requested):

inst/include/dust/dust.hpp
richfitz: on line 83, no?

johnlees: Yes, I think was just obscured by the diff!

johnlees (approved):

PR #206, opened 2021-03-17
Allow thowing with impossible rbinom call

### LINKED ISSUE #190: Exception for invalid rbinom calls
Currently we have a comment:
```
    // TODO: Should control for this too, but not really clear what we
    // need to do to safely deal.
    /*
      if (n < 0 || p < 0 || p > 1) {
      return NaN;
      }
    */
```
replacing this with
```
if (n < 0 || p < 0 || p > 1) {
      real_t null = NAN;
      return null;
}
```
will run, but is probably an error

With CUDA, we can use assert:
```
#include <assert.h>
...
if (n < 0 || p < 0 || p > 1) {
     printf("Invalid call to rbinom with n = %d p = %f\n", n, p);
     assert(0);
}
```
with openmp we need a try/except, see [guide](https://docs.microsoft.com/en-us/cpp/parallel/concrt/convert-an-openmp-loop-that-uses-exception-handling?view=msvc-160)

richfitz:
this is tricky only because we need to safely throw from the openmp block. For that I have made a little utility class that safely collects errors and can rethrow them as a group. That then leaves the object inconsistent so there's a little faff here to make sure the user can't continue.

With this in place, model logic errors will throw rather than leave the model running forever.

Fixes #190


NEWS.md
```diff
@@ -1,3 +1,7 @@
+# dust 0.8.7
+
+* Invalid inputs in `rbinom` are converted into exceptions which are safely thrown even from parallel code. After an error, the object will be in an inconsistent state as not all particles will have completed their run and you must reset the object (via `reset()` or `set_state()`) to continue (#190)
```
johnlees, line 3: The second sentence should probably be in docs somewhere, rather than/as well as news

inst/include/dust/utils.hpp
```diff
@@ -64,6 +64,15 @@ HOSTDEVICE T epsilon() {
 #endif
 }
 
+inline void throw_message(const char * msg) {
```
johnlees, line 67: This needs to be `HOSTDEVICE`

inst/include/dust/utils.hpp
```diff
@@ -64,6 +64,15 @@ HOSTDEVICE T epsilon() {
...
+#ifdef __CUDA_ARCH__
+    printf("%s\n", msg);
+    assert(0);
```
johnlees, line 70: Add `#include <assert.h>` at the top of the file

inst/include/dust/types.hpp
```diff
@@ -291,6 +295,59 @@ T* shared_copy(T* dest, const T src) {
...
+    return count > 0;
+  }
+
+  template <typename T>
+  void capture(const T& e, size_t i) {
+    err[i] = e.what();
+    seen[i] = true;
+  }
+
+  void report(size_t n_max = 4) {
```
johnlees, line 322: Note that only the top four are being reported in the message?

johnlees (changes requested):

inst/include/dust/types.hpp
richfitz: that is reported at the moment:

```
Error: 10 particles reported errors.
  - 1: Invalid call to rbinom with n = -10, p = 0.0237906
  - 2: Invalid call to rbinom with n = -10, p = 0.0237906
  - 3: Invalid call to rbinom with n = -10, p = 0.0237906
  - 4: Invalid call to rbinom with n = -10, p = 0.0237906
  - (and 6 more)
```

if you have fewer than 4 it looks like:

```
Error: 2 particles reported errors.
  - 4: Invalid call to rbinom with n = -1, p = 0.0237906
  - 9: Invalid call to rbinom with n = -10, p = 0.0237906
```

richfitz: I wondered about making this configurable but honestly I think that's more drama than it's worth. The most likely situation seems to be we get one or two (which this will cope with nicely and give a reasonable length error) or they *all* die in which case we don't want to have a billion lines printed. Probably on error the user will have to get involved with gdb to really progress anyway

NEWS.md
richfitz: good shout, updated the docs (though this also makes it clear they need an overhaul - see #207)

johnlees (approved):

Issue #177, opened 2021-02-25
Filter should allow partial running


richfitz:
Required for supporting SMC^2 in mcstate:

* run to some step index
* ability to fork
* return partial log likelihood

PR #214, opened 2021-03-22
Move filter into its own file

### LINKED ISSUE #213: Move filter code into its own file
To help with #154, the filter code can be rewritten as a free function

richfitz:
Move the filter code into its own file. This will help with designing logic to help with the partial run filter (#177) and reduces the number of things that the Dust object holds

Fixes #213

### DIFF
--- a/inst/include/dust/filter.hpp
+++ b/inst/include/dust/filter.hpp
@@ -4,7 +4,7 @@
 namespace dust {
 
 template <typename T>
-std::vector<typename T::real_t> filter(T * obj,
+std::vector<typename T::real_t> filter(Dust<T> * obj,
                                        filter_state<typename T::real_t>& state,
                                        bool save_trajectories,
                                        std::vector<size_t> step_snapshot) {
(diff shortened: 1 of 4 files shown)


inst/include/dust/filter.hpp
```diff
@@ -0,0 +1,73 @@
+#ifndef DUST_FILTER_HPP
+#define DUST_FILTER_HPP
+
+namespace dust {
+
+template <typename T>
+std::vector<typename T::real_t> filter(T * obj,
```
johnlees, line 7: Would `Dust<T>` be a better template?

johnlees: Or does that make getting real_t annoying. Just thinking that obj feels a bit too general

johnlees (approved):

inst/include/dust/filter.hpp
richfitz: hmm, good shout, will check

PR #216, opened 2021-03-22
Allow running all rng functions with floats

### LINKED ISSUE #191: Proper tests for all float rng calculations
Expose this via the rng interface for better testing

richfitz:
This PR allows running all the rngs in float mode. This will hopefully make corner cases easier to work with as we hit them

Fixes #191

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: dust
 Title: Iterate Multiple Realisations of Stochastic Models
-Version: 0.8.11
+Version: 0.8.12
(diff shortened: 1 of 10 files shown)


johnlees (commented):
This all looks good, but I wonder whether we should follow a similar approach to the dust real_t definition where we give the type as a free input, rather than a boolean between float/double. This might give us more flexibility in future in case we ever want to use say fp16, or some of the more exotic mixed precision types. What do you reckon?

johnlees (approved):

PR #219, opened 2021-03-22
Split help into interface/reference

### LINKED ISSUE #207: Restructure help
Currently interface.R contains the help for compiling dust objects and unrelated help for driving dust objects. The latter we want to make available to other packages easily

richfitz:
This PR restructures help so that it's easier to refer to reference for using a dust object without having to link to the page that creates a dust object

Fixes #207


johnlees (approved):
Do the package versions of dust models always need a `zzz.R`? If so, could we add/explain that in the help?

richfitz:
> Do the package versions of dust models always need a zzz.R? If so, could we add/explain that in the help?

They don't need a zzz.R, just some way of loading the dynamic lib - I think that's not actually noted anywhere though, will add a note

PR #223, opened 2021-03-24
Make rpois work on device

richfitz:
Couple of tweaks here, compilation works on Raphael's branch of sircovid but that's about it.

### DIFF
--- a/inst/include/dust/distr/poisson.hpp
+++ b/inst/include/dust/distr/poisson.hpp
@@ -2,10 +2,12 @@
 #define DUST_DISTR_POISSON_HPP
 
 #include <cmath>
+#include <dust/utils.hpp>
 
 namespace dust {
 namespace distr {
 
+__nv_exec_check_disable__
 template <typename real_t>
 HOSTDEVICE int rpois(rng_state_t<real_t>& rng_state,
                      typename rng_state_t<real_t>::real_t lambda) {
@@ -30,8 +32,7 @@ HOSTDEVICE int rpois(rng_state_t<real_t>& rng_state,
     while (true) {
       real_t u = dust::unif_rand<real_t>(rng_state);
       prod = prod * u;
-      if (prod <= exp_neg_rate &&
-          x <= std::numeric_limits<int>::max()) {
+      if (prod <= exp_neg_rate && x <= dust::utils::integer_max()) {
         break;
       }
       x++;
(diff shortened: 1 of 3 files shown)


inst/include/dust/utils.hpp
```diff
@@ -64,6 +64,37 @@ HOSTDEVICE T epsilon() {
 #endif
 }
 
+__nv_exec_check_disable__
+inline int integer_max() {
```
johnlees, line 68: This needs to be `HOSTDEVICE`

inst/include/dust/utils.hpp
```diff
@@ -64,6 +64,37 @@ HOSTDEVICE T epsilon() {
...
+#else
+  return std::numeric_limits<int>::max();
+#endif
+}
+
+template <typename T>
+T lgamma(T x);
+
+// We need this for the lgamma in rpois to work
+#ifdef __CUDA_ARCH__
```
johnlees, line 80: I don't think this ifdef is needed, you should just be able to mark the functions with DEVICE and HOST (which you have already done – except HOST is implicit, but I think it helps to make it explicit in this case)

johnlees: Also I think this ifdef will never be true

inst/include/dust/utils.hpp
```diff
@@ -64,6 +64,37 @@ HOSTDEVICE T epsilon() {
...
+inline int integer_max() {
+#ifdef __CUDA_ARCH__
+  return INT_MAX;
+#else
+  return std::numeric_limits<int>::max();
+#endif
+}
+
+template <typename T>
+T lgamma(T x);
```
johnlees, line 77: Not sure how host/device works in this case – is this prototype needed?
I think that this: https://github.com/mrc-ide/dust/blob/master/inst/include/dust/xoshiro.hpp#L142-L169
is the closest we have to this logic in the existing code

inst/include/dust/distr/poisson.hpp
```diff
@@ -102,7 +103,7 @@ int rpois(rng_state_t<real_t>& rng_state,
       // The expression below is equivalent to the computation of step 2)
       // in transformed rejection (v <= alpha * F'(G(u)) * G'(u)).
       real_t s = std::log(v * inv_alpha / (a / (u_shifted * u_shifted) + b));
-      real_t t = -lambda + k * log_rate - std::lgamma(k + 1);
+      real_t t = -lambda + k * log_rate - dust::utils::lgamma<real_t>(k + 1);
```
johnlees, line 106: Would it make sense to update the lgamma calls in the densities now too? Happy for this to be left for a future PR also

johnlees (changes requested):

johnlees (approved):

PR #224, opened 2021-03-25
CUDA version of whole particle filter

johnlees:
TODO:
- [x] Save state in a device history object
- [x] Write a kernel to extract this history in the right order
- [x] Save full state snapshots in a device snapshot object
- [x] Write a kernel to extract these snapshots in the right order
- [x] Flatten the data object and copy to device
- [x] Make copy into shared memory a function to be shared by run_device and compare_device
- [x] Write a compare_device kernel
- [x] Write a scale_log_weights kernel which uses cub::DeviceSegmentedReduce::Max and an exp kernel
- [x] Write a resample kernel which uses a prefix sum and interval bisection (per pars), then calls reorder
- [x] Add an `ALIGN` directive to `data_t`
- [x] ~(optional) move `u` generation for resample to device, depending on profile~

style:
- [x] Reduce duplication in the filter history objects
- [x] Add a stale check on device_select runs
- [x] struct for new device_array objects needed for filter (data, weights etc)
- [x] (after profiling) look for opportunities to async compute/copy
- [x] Set an extra RNG state for host generation (at the end of particles? at the start?)

testing:
- [x] Update the CI tests
- [x] Results are the same for single parameter set
- [x] Results are the same for multiple parameter sets
- [ ] Check whether doubles are a better choice for filter kernels

issues:
- [x] Saving history order is causing a cuda malloc & free
- [x] Long gap spent in host after run_particles (doing set_step on every particle?)


inst/include/dust/device_state.cuh
```diff
@@ -0,0 +1,78 @@
...
+template <typename T>
+DEVICE dust::device_ptrs<T>
+load_shared_state(const int pars_idx,
+                  const size_t n_shared_int,
+                  const size_t n_shared_real,
+                  const int * shared_int,
+                  const typename T::real_t * shared_real,
+                  const typename T::data_t * data,
+                  bool use_shared_L1) {
+  // Particle index i, and max index to process in the block
```
richfitz, line 19: What does this comment refer to?

inst/include/dust/device_state.cuh
```diff
@@ -0,0 +1,78 @@
+// -*- c++ -*-
+#ifndef DUST_DEVICE_STATE_CUH
+#define DUST_DEVICE_STATE_CUH
+
+#include <dust/cuda.cuh>
+#include <dust/types.hpp>
+
+namespace dust {
+
+template <typename T>
```
richfitz, line 10: This really begs for a comment outlining what this is doing at a high level

inst/include/dust/dust.hpp
```diff
@@ -308,6 +310,19 @@ class Dust {
     }
   }
 
+  // Used for copy of state into another block of memory on the device (used
+  // for history saving)
+  void state(dust::device_array<real_t>& device_state, const size_t dst_offset,
```
richfitz, line 315: I may fail to find where this is called or connect my thoughts up here, but:

Is there a good reason this overloads the state name? It seems more like something that is more connected to the device logic and better suited to being a private method with a different name

inst/include/dust/dust.hpp
```diff
@@ -336,6 +351,12 @@ class Dust {
     }
   }
 
+  void state_full(dust::device_array<real_t>& device_state, size_t dst_offset) {
```
richfitz, line 354: As state() above, not sure this is a public overload of state_full

inst/include/dust/dust.hpp
```diff
@@ -405,26 +427,117 @@ class Dust {
     if (n_pars_ == 0) {
       // One parameter set; shuffle among all particles
       const size_t np = particles_.size();
-      real_t u = dust::unif_rand(rng_.state(0));
+      real_t u = dust::unif_rand(rng_.state(n_particles_total_));
```
richfitz, line 430: nice

inst/include/dust/dust.hpp
```diff
@@ -405,26 +427,117 @@ class Dust {
...
+#ifdef __NVCC__
+    // Cumulative sum
+    // Note this is over all parameters, this is fixed with a
+    // subtraction in the interval kernel
+    cub::DeviceScan::InclusiveSum(scan.scan_tmp.data(),
+                                  scan.tmp_bytes,
+                                  weights.data(),
+                                  scan.cum_weights.data(),
+                                  scan.cum_weights.size());
+    // Don't sync here
```
richfitz, line 465: can you add a reason?

inst/include/dust/dust.hpp
```diff
@@ -405,26 +427,117 @@ class Dust {
...
-        dust::filter::resample_weight(it_weights + j, np, u, j, it_index + j);
+        dust::filter::resample_weight(it_weights + j, np, u[i], j, it_index + j);
       }
     }
 
     reorder(index);
     stale_device_ = true;
   }
 
+  void resample_device(dust::device_array<real_t>& weights,
```
richfitz, line 453: This method is pretty big - can the implementation be farmed off into a free function?

inst/include/dust/dust.hpp
```diff
@@ -521,6 +640,101 @@ class Dust {
...
+
+  template <typename U = T>
+  typename std::enable_if<dust::has_gpu_support<U>::value, void>::type
+  compare_data_device(dust::device_array<real_t>& res,
+                      const size_t data_offset) {
+    refresh_device();
+#ifdef __NVCC__
+    const int warp_size = dust::cuda::warp_size;
+    // Set up blocks and shared memory preferences
+    size_t blockSize = 128;
```
richfitz, line 677: this appears in enough places we should make it a `#define` in our cuda header?

inst/include/dust/filter.hpp
```diff
@@ -72,6 +72,72 @@ std::vector<typename T::real_t> filter(Dust<T> * obj,
...
+
+  std::vector<real_t> ll_host(n_pars, 0);
+  dust::device_array<real_t> log_likelihood(ll_host);
+  dust::device_weights<real_t> weights(n_particles, n_pars);
+  dust::device_scan_state<real_t> scan;
+  scan.initialise(n_particles, weights.weights());
+
+  if (save_trajectories) {
+    state.trajectories.resize(obj->n_state(), n_particles, n_data);
+    obj->state(state.trajectories.values(), state.trajectories.value_offset());
```
richfitz, line 95: I see this is where state is used - this does look very similar to how the other state() interfaces look.

If we could do this as:

```
    obj->state(state.trajectories.values() + state.trajectories.value_offset());
```

I think it would be closer

inst/include/dust/filter_kernels.hpp
```diff
@@ -0,0 +1,43 @@
...
+#endif
+    real_t ll_scaled = std::log(ll_step[i] / n_particles_each) + max_weights[i];
+    ll_step[i] = ll_scaled;
+    ll[i] += ll_scaled;
+  }
+}
+
+}
+
+#endif
```
richfitz, line 43: can you add the final newline here-  it causes warnings in a few things not to have it

inst/include/dust/filter_state.hpp
```diff
@@ -80,74 +66,245 @@ class filter_trajectories {
...
+    return this->offset_ * this->n_state_ * this->n_particles_;
+  }
+
+  template <typename Iterator>
+  void history(Iterator dest) const {
+    // Copy from D->H
+    std::vector<real_t> state_host(size());
+    state_.get_array(state_host);
+    // Destride and copy into iterator
+    // TODO: openmp here? collapse(2)
```
richfitz, line 279: does this show up in the profile at all?

inst/include/dust/interface.hpp
```diff
@@ -402,34 +435,31 @@ cpp11::sexp dust_filter(SEXP ptr, bool save_trajectories,
...
-
-  if (r_step_snapshot != R_NilValue) {
-    const auto& snapshots = filter_state.snapshots;
-    cpp11::writable::doubles snapshots_data(snapshots.size());
-    snapshots.history(REAL(snapshots_data));
-    snapshots_data.attr("dim") =
-      dust::interface::state_array_dim(obj->n_state_full(), obj->shape(),
-                                       step_snapshot.size());
-    r_snapshots = snapshots_data;
+  if (device) {
```
richfitz, line 443: this might be clearer if the identical inner parts are written as a free function templated over the filter state

richfitz (changes requested):
I think you'll have to talk me through this at a high level, but the code itself is largely pretty tidy considering the awfulness of CUDA.

We should aim to get a compilation test into the docker/cuda test (docker/compile_cuda_model)

Have you managed to test this on sircovid?

inst/examples/sirs.cpp
```diff
@@ -1,9 +1,12 @@
 class sirs {
 public:
-  typedef double real_t;
-  typedef dust::no_data data_t;
+  typedef float real_t;
```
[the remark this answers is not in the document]
johnlees: Changed to a double

inst/include/dust/containers.hpp
```diff
@@ -103,25 +105,50 @@ class device_array {
...
+    } else {
+      CUDA_CALL(cudaMemcpy(data_, src.data(), size_ * sizeof(T),
+                          cudaMemcpyDefault));
+    }
 #else
     std::memcpy(data_, src.data(), size_ * sizeof(T));
 #endif
   }
 
+  void set_array(const T* src, const size_t src_size,
```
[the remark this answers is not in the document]
johnlees: Added comments to clarify

inst/include/dust/densities.hpp
```diff
@@ -35,37 +53,45 @@ T dbinom(int x, int size, T prob, bool log) {
...
 template <typename T>
-T dnorm(T x, T mu, T sd, bool log) {
+HOSTDEVICE T dnorm(T x, T mu, T sd, bool log) {
   if (sd == 0) {
     return ddelta(x - mu, log);
   }
+#ifdef __CUDA_ARCH__
+  const T m_ln_sqrt_2pi = static_cast<T>(const_m_ln_sqrt_2pi);
+#else
   constexpr T m_ln_sqrt_2pi = 0.918938533204672741780329736406;
```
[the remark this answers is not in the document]
johnlees: Changed this to be the same as gamma_tail

inst/include/dust/dust.hpp
johnlees: Going to leave these for now. Once we've done #227 I think we should be able to combine the two filter functions with a template by using this style

inst/include/dust/dust.hpp
johnlees: Although 128 is most common, this is generally different for each kernel, and the result of optimising with the profiler. I will give all of these different names however, to make this clearer

inst/include/dust/filter.hpp
johnlees: Leaving until #227 when this will change again

inst/include/dust/filter_state.hpp
johnlees: Not sure, we'll have to check with sircovid. Options to deal with this if it is slow would be to either:
- Add an openmp block and see if it helps
- Additionally use std::async or similar to compute this as the next step of the filter is running
- Write a CUDA version of this (similar to the scatter kernel). This would be nicest combined with the particle_ancestry function, which we may want to redo so it's not run backwards after we have the final step (if that's possible)

inst/include/dust/densities.hpp
```diff
@@ -35,37 +53,45 @@ T dbinom(int x, int size, T prob, bool log) {
...
   const T dx = x - mu;
   const T ret = - dx * dx / (2 * sd * sd) - m_ln_sqrt_2pi - std::log(sd);
   return maybe_log(ret, log);
 }
 
 template <typename T>
-T dnbinom(int x, T size, T mu, bool log) {
+HOSTDEVICE T dnbinom(int x, T size, T mu, bool log) {
+#ifndef __CUDA_ARCH__
+  // static_assert(!std::is_floating_point<T>::value,
```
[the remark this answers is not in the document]
johnlees: These should be true not false, doh

inst/include/dust/dust.hpp
```diff
@@ -521,6 +540,72 @@ class Dust {
     for (size_t i = 0; i < particles_.size(); ++i) {
       res[i] = particles_[i].compare_data(data[i / np], rng_.state(i));
     }
+    stale_device_ = true; // RNG use
+  }
+
+  std::vector<real_t> compare_data_device() {
+    std::vector<real_t> res;
```
richfitz, line 547: This requires a device refresh

richfitz (changes requested):

inst/include/dust/dust.hpp
johnlees: I think that's done on L573 when `compare_data_device(x,x)` is called from this function

johnlees: fixed in 30f9eb8

richfitz (approved):

PR #234, opened 2021-05-18
Allow changing real_t size at compilation

### LINKED ISSUE #233: Add helper to automatically rewrite the real_t on gpu compilation

richfitz:
Adds a little helper option to allow changing the real_t. Typically we won't use this because odin.dust does this much better. However, it'll be useful for benchmark/docs where we want gpu versions of models in float mode without much weirdness.

Also include `sizeof(T::real_t)` in the device info. Seems an ok place to keep it.

I've also tidied up the sirs example a bit and exposed more parameters that we'll need in the bechmarking

Fixes #233


R/interface.R
```diff
@@ -242,6 +256,20 @@ dust_workdir <- function(path) {
 }
 
 
+dust_rewrite_real <- function(filename, real_t) {
```
johnlees, line 259: Should we check that this is either double or float?

johnlees (approved):

R/interface.R
richfitz: I wondered but also figured the compiler would whinge. This also allows us later on to fiddle with shorts etc.

PR #238, opened 2021-06-09
More robust resampling

### LINKED ISSUE #237: memory error with particle filter and float models

richfitz:
This PR fixes an issue in the resampling when using `float` (rather than `double`). Only seemed to affect things when we drew a very large `u`.

See https://github.com/reside-ic/reside-ic.github.io/pull/60 for details

Fixes #237

### DIFF
--- a/inst/include/dust/filter_tools.hpp
+++ b/inst/include/dust/filter_tools.hpp
@@ -14,14 +14,15 @@ void resample_weight(typename std::vector<real_t>::const_iterator w,
   const real_t tot = std::accumulate(w, w + n, static_cast<real_t>(0));
   real_t ww = 0.0, uu0 = tot * u / n, du = tot / n;
   size_t j = offset;
+  const size_t end = n + offset;
   for (size_t i = 0; i < n; ++i) {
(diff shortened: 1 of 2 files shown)


johnlees (approved):

inst/include/dust/filter_tools.hpp
```diff
@@ -12,16 +12,21 @@ void resample_weight(typename std::vector<real_t>::const_iterator w,
...
+  real_t ww = 0.0, uu0 = tot * u / n, du = tot / n;
   size_t j = offset;
+  const size_t end = n + offset;
   for (size_t i = 0; i < n; ++i) {
-    while (ww < uu) {
+    // We could accumulate uu by adding du at each iteration but that
+    // suffers roundoff error here with floats.
+    const real_t uu = uu0 + i * du;
+    // The second clause (i.e., j - offset < n) should never be hit
+    // but prevents any invalid read if we have pathalogical 'u' that
```
giovannic, line 23:
```suggestion
    // but prevents any invalid read if we have pathological 'u' that
```

PR #240, opened 2021-06-18
Avoid precision loss in dnbinom

### LINKED ISSUE #239: Precision loss with floats in dnbinom

richfitz:
Seen while running sircovid in float mode on real data, where we hit a point with zero modelled + zero observed deaths.

Fixes #239

### DIFF
--- a/inst/include/dust/densities.hpp
+++ b/inst/include/dust/densities.hpp
@@ -81,7 +81,6 @@ HOSTDEVICE T dnbinom(int x, T size, T mu, bool log) {
   static_assert(std::is_floating_point<T>::value,
                 "dnbinom should only be used with real types");
 #endif
-  const T prob = size / (size + mu);
   if (x == 0 && size == 0) {
     return maybe_log(0, log);
   }
@@ -91,11 +90,24 @@ HOSTDEVICE T dnbinom(int x, T size, T mu, bool log) {
   if (mu == 0) {
     return maybe_log(x == 0 ? 0 : -dust::utils::infinity<T>(), log);
   }
-  const T ret = dust::utils::lgamma(static_cast<T>(x + size)) -
-    dust::utils::lgamma(static_cast<T>(size)) -
-    dust::utils::lgamma(static_cast<T>(x + 1)) +
-    size * std::log(prob) + x * std::log(1 - prob);
-  return maybe_log(ret, log);
+
+  // Avoid size / (size + mu) where this would cause prob to be
+  // equal. Somewhat arbitrarily, taking 100 * floating point eps as
+  // the change over.
+  const T ratio = dust::utils::epsilon<T>() * 100;
+  if (mu < ratio * size) {
+    const T p = std::log(mu / (1 + mu / size));
+    const T ret = x * p - mu - lgamma(x + 1) +
+      std::log1p(x * (x - 1) / (2 * size));
+    return maybe_log(ret, log);
+  } else {
+    const T prob = size / (size + mu);
+    const T ret = dust::utils::lgamma(static_cast<T>(x + size)) -
+      dust::utils::lgamma(static_cast<T>(size)) -
+      dust::utils::lgamma(static_cast<T>(x + 1)) +
+      size * std::log(prob) + x * std::log(1 - prob);
+    return maybe_log(ret, log);
+  }
 }
 
 // A note on this parametrisation:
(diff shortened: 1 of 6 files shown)


src/densities.cpp
```diff
@@ -24,17 +24,26 @@ SEXP dust_dnorm(cpp11::doubles x, cpp11::doubles mu, cpp11::doubles sd,
...
   }
   return ret;
 }
 
+
+[[cpp11::register]]
+SEXP dust_dnbinom(cpp11::integers x, cpp11::doubles size, cpp11::doubles mu,
+                  bool log, bool is_float) {
+  return is_float ?
+    dust_dnbinom_<float>(x, size, mu, log) :
```
johnlees, line 43: Note that the `is_float` is used primarily for testing

tests/testthat/test-densities.R
```diff
@@ -52,64 +52,83 @@ test_that("dbinom agrees", {
...
-    dnbinom(511, 2, mu = 6.65, log = TRUE),
-    dust_dnbinom(511L, 2, 6.65, TRUE))
-
-  ## Allow non integer size (wrong in <= 0.7.5)
-  expect_equal(
-    dust_dnbinom(x = 511L, size = 3.5, mu = 1, log = TRUE),
-    dnbinom(511, 3.5, mu = 1, log = TRUE))
+  for (is_float in c(FALSE, TRUE)) {
+    if (is_float) {
+      tolerance <- sqrt(sqrt(.Machine$double.eps))
```
johnlees, line 57: Why sqrt twice?

johnlees: I might be getting this wrong but I find that:
sqrt(flt_epsilon) = sqrt(2^-23) = 0.0003449638
sqrt(sqrt(dbl_epsilon) = sqrt(sqrt(2^-52)) = 0.0001220703

inst/include/dust/densities.hpp
```diff
@@ -91,11 +90,24 @@ HOSTDEVICE T dnbinom(int x, T size, T mu, bool log) {
...
   }
-  const T ret = dust::utils::lgamma(static_cast<T>(x + size)) -
-    dust::utils::lgamma(static_cast<T>(size)) -
-    dust::utils::lgamma(static_cast<T>(x + 1)) +
-    size * std::log(prob) + x * std::log(1 - prob);
-  return maybe_log(ret, log);
+
+  // Avoid size / (size + mu) where this would cause prob to be
+  // equal. Somewhat arbitrarily, taking 100 * floating point eps as
+  // the change over.
```
johnlees, line 96:
```suggestion
  // Avoid size / (size + mu) when size is close to zero, and this would cause prob to be
  // equal to zero. Somewhat arbitrarily, taking 100 * floating point eps as
  // the change over.
```

inst/include/dust/densities.hpp
```diff
@@ -91,11 +90,24 @@ HOSTDEVICE T dnbinom(int x, T size, T mu, bool log) {
...
+  const T ratio = dust::utils::epsilon<T>() * 100;
+  if (mu < ratio * size) {
+    const T p = std::log(mu / (1 + mu / size));
+    const T ret = x * p - mu - lgamma(x + 1) +
+      std::log1p(x * (x - 1) / (2 * size));
+    return maybe_log(ret, log);
```
johnlees, line 102:
```suggestion
    const T log_prob = std::log(mu / (1 + mu / size));
    const T ret = x * log_prob - mu - lgamma(x + 1) +
      std::log1p(x * (x - 1) / (2 * size));
    return maybe_log(ret, log);
```

johnlees: Where does this formulation come from, I couldn't see how the `ret = ...` formula follows from size =~ 0

johnlees (changes requested):

src/densities.cpp
richfitz: All the R versions of the density functions are only used for testing

tests/testthat/test-densities.R
richfitz: flt_eps is ~1e-8 (https://en.wikipedia.org/wiki/Machine_epsilon), double is ~1e-16

johnlees: Seems fine to round those, but I don't understand the rationale for writing the approximation as flt_eps =~ sqrt(dbl_eps) which I understand as being roughly true by chance, rather than just writing flt_eps as a constant. Is it expected to be machine/implementation specific?

johnlees: (this doesn't matter for the code though, just asking out of interest)

richfitz: just avoiding magic numbers and writing everything in terms of eps :)

johnlees (approved):

PR #243, opened 2021-06-21
Rejoin after divergences in kernel functions

johnlees:
This was mostly checked in rbinom before, but this applies those findings across the r<dist> and d<dist> functions. This is only needed with branching (if/break/continue/until) in KERNEL, DEVICE or HOSTDEVICE functions, so I think this covers all of those.

### DIFF
--- a/inst/include/dust/densities.hpp
+++ b/inst/include/dust/densities.hpp
@@ -75,16 +75,17 @@ template <typename T>
 HOSTDEVICE T dnorm(T x, T mu, T sd, bool log) {
   T ret;
   if (sd == 0) {
-    ret = ddelta(x - mu, log);
+    ret = ddelta(x - mu, log); // This does maybe_log
   } else {
     const T dx = x - mu;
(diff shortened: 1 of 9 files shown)


inst/include/dust/densities.hpp
```diff
@@ -50,12 +50,18 @@ HOSTDEVICE T dbinom(int x, int size, T prob, bool log) {
...
+  } else {
+    ret = lchoose<T>(size, x) +
+      x * std::log(prob) +
+      (size - x) * std::log(1 - prob);
   }
-  const T ret = lchoose<T>(size, x) +
-    x * std::log(prob) +
-    (size - x) * std::log(1 - prob);
+
+#ifdef __CUDA_ARCH__
```
richfitz, line 62: Could we put this into `inst/include/dust/cuda.cuh` so that we only see here

```
SYNCWARP
```

perhaps?

richfitz (changes requested):

richfitz (approved):

PR #246, opened 2021-06-23
Enable keeping integers in shared memory, even if int + shared is too large

richfitz:
This PR tries to make the shared memory handling a bit more clever by keeping integers in shared memory even where reals don't fit.

We can make this more clever again later by copying across some fraction of reals too, but this will hopefully help by keeping offsets and dimensions handy as they're both heavily used.

I've factored out the fairly nasty calculations into their own classes and exposed that via a wrapper function for testing

~I've run this very quickly in sircovid, and while it works (i.e., it runs through) it now takes considerably longer than before...~

Timings on my 2080 with the two dust versions, no difference at all :thinking: 

```
## dust 0.9.8
##  1600: 23.9
##  3200: 24.5
##  6400: 27.4
## 12800: 48.7

## dust 0.9.9
##  1600: 23.2
##  3200: 24.6
##  6400: 27.7
## 12800: 48.7
```


inst/include/dust/cuda_launch_control.hpp
```diff
@@ -0,0 +1,233 @@
...
+    enabled_(device_id_ >= 0) {
+#ifdef __NVCC__
+    if (device_id_ >= 0) {
+      CUDA_CALL(cudaSetDevice(device_id_));
+      CUDA_CALL(cudaDeviceSetCacheConfig(cudaFuncCachePreferL1));
+    }
+#endif
+  }
+
+  const int device_id_;
```
johnlees, line 48:
```suggestion
private:
  const int device_id_;
```

johnlees: or this is a struct not a class?

inst/include/dust/cuda_launch_control.hpp
```diff
@@ -0,0 +1,233 @@
...
+#ifdef DUST_USING_CUDA_PROFILER
+  if (config.enabled_) {
+      std::cout << "Stopping profiler" << std::endl;
+      CUDA_CALL(cudaProfilerStop());
+  }
+#endif
+}
+
+
+class launch_control_dust {
```
johnlees, line 74: Also a struct?

inst/cuda/dust.cu
```diff
@@ -3,5 +3,5 @@
 #include <dust/device_info.hpp>
 
 cpp11::sexp dust_device_info() {
-  return dust::cuda::device_info<void>();
+  return dust::cuda::device_info<float>();
```
johnlees, line 6: Is it correct that this is always float? Might it sometimes be double

inst/include/dust/cuda_launch_control.hpp
```diff
@@ -0,0 +1,233 @@
...
+                               n_shared_int, n_shared_real,
+                               real_size, 0, shared_size, run_block_size);
+    compare = launch_control_model(n_particles, n_particles_each,
+                                   n_shared_int, n_shared_real,
+                                   real_size, data_size, shared_size, 128);
+
+    reorder       = launch_control_simple(128, n_particles * n_state_full);
+    scatter       = launch_control_simple( 64, n_particles * n_state_full);
+    index_scatter = launch_control_simple( 64, n_particles * n_state);
+    interval      = launch_control_simple(128, n_particles);
```
johnlees, line 218: Maybe we should make the default block sizes const at the top of the file so that they are easier/clearer to change in future profiling?

johnlees: Although that being said it's quite clear writing them out here too

inst/include/dust/dust.hpp
```diff
@@ -32,48 +33,43 @@ class Dust {
...
-#ifdef __NVCC__
-    initialise_device(device_id);
-#endif
+    select_scatter_(false) {
     initialise(pars, step, n_particles, true);
     initialise_index();
     shape_ = {n_particles};
+#ifdef DUST_USING_CUDA_PROFILER
+    cuda_profiler_start(device_config_);
+#endif
```
johnlees, line 54: These ifdefs are also used inside the function, I don't think they are necessary here too

inst/include/dust/dust.hpp
```diff
@@ -32,48 +33,43 @@ class Dust {
...
   }
 
   Dust(const std::vector<pars_t>& pars, const size_t step,
        const size_t n_particles, const size_t n_threads,
-       const std::vector<uint64_t>& seed, size_t device_id,
+       const std::vector<uint64_t>& seed, cuda::device_config device_config,
```
johnlees, line 59:
```suggestion
       const std::vector<uint64_t>& seed, cuda::device_config& device_config,
```

inst/include/dust/dust.hpp
```diff
@@ -32,48 +33,43 @@ class Dust {
 
   Dust(const pars_t& pars, const size_t step, const size_t n_particles,
        const size_t n_threads, const std::vector<uint64_t>& seed,
-       size_t device_id) :
+       cuda::device_config device_config) :
```
johnlees, line 36:
```suggestion
       cuda::device_config& device_config) :
```

inst/include/dust/dust.hpp
```diff
@@ -83,14 +79,17 @@ class Dust {
     for (auto i : shape) {
       shape_.push_back(i);
     }
+#ifdef DUST_USING_CUDA_PROFILER
+    cuda_profiler_start(device_config_);
+#endif
```
johnlees, line 84: unnecessary ifdefs

inst/include/dust/dust.hpp
```diff
@@ -83,14 +79,17 @@ class Dust {
...
   }
 
   // We only need a destructor when running with cuda profiling; don't
   // include ond otherwise because we don't actually follow the rule
   // of 3/5/0
-#ifdef DUST_ENABLE_CUDA_PROFILER
+#ifdef DUST_USING_CUDA_PROFILER
```
johnlees, line 90: Also unnecessary

inst/include/dust/dust.hpp
```diff
@@ -863,17 +837,22 @@ class Dust {
...
     select_needed_ = true;
     if (!std::is_sorted(index_.cbegin(), index_.cend())) {
       select_scatter_ = true;
     } else {
       select_scatter_ = false;
     }
+
+    // TODO: get this 64 from the original configuration, if possible.
+    cuda_pars_.index_scatter =
+      dust::cuda::launch_control_simple(64, n_particles * n_state());
```
johnlees, line 855: If we had the defaults as consts we could use the variable to replace 64 here too

inst/include/dust/interface_helpers.hpp
```diff
@@ -363,6 +364,33 @@ std::vector<size_t> check_step_snapshot(cpp11::sexp r_step_snapshot,
...
 
+
+inline dust::cuda::device_config device_config(cpp11::sexp r_device_config) {
+  size_t run_block_size = 128;
+
+  cpp11::sexp r_device_id = r_device_config;
+  if (TYPEOF(r_device_config) == VECSXP) {
+    cpp11::list r_device_config_l = cpp11::as_cpp<cpp11::list>(r_device_config);
+    r_device_id = r_device_config_l["device_id"]; // could error if missing?
+    cpp11::sexp r_run_block_size = r_device_config_l["run_block_size"];
```
johnlees, line 375: Do we check this is a multiple of the warp size somewhere? (not an error, but definitely not a good idea if not % 32)

johnlees (changes requested):

inst/include/dust/cuda_launch_control.hpp
richfitz: we refer to these later - it's basically a struct with a nontrivial constructor (the only difference between a struct and a class is the default visibility of the members, really)

inst/include/dust/cuda_launch_control.hpp
richfitz: as above, structy-class or classy-struct

inst/cuda/dust.cu
richfitz: this is the default value printed in the information, so the size of floats is nicer to see here than the confusing zero. This is purely informational

inst/include/dust/dust.hpp
richfitz: This won't change the number of copies (we could pass it as `const cuda::device_config&` but at some point it will get copied I belive

inst/include/dust/dust.hpp
richfitz: This one is necessary, as we don't generally want the destructor - I've put the other ones in for symmetry, but it still feels pretty nasty. I did wonder about putting the profile control into the control object but had issues with the destructor there too...

inst/include/dust/interface_helpers.hpp
richfitz: I'm fine with making this an error, will do that now

inst/include/dust/dust.hpp
richfitz: after some back and forth, decided to keep this in so the destructor remains clear

johnlees (approved):

PR #251, opened 2021-07-29
Support for running models in "deterministic" mode

### LINKED ISSUE #250: Run in deterministic mode
Requires:

* support in each distribution
* switch/flag in rng state
* control in run method

richfitz:
This PR makes several changes to the rng engine to allow stochasticity to be turned off:

* rng_state gets a flag indicating if it's enabled
* each distribution function inspects that flag on the cpu, we return the mean in this case (e.g., `n * p` for rbinom). *this changes the return type to `real_t`, which will have little practical consequences for us*
* the `run` function gets a `deterministic` argument, which disables the RNG, runs the steps, then re-enables the RNG

Fixes #250


inst/include/dust/distr/binomial.hpp
```diff
@@ -166,10 +166,12 @@ inline HOSTDEVICE real_t btrs(rng_state_t<real_t>& rng_state, int n_int, real_t
   return draw;
 }
 
+// NOTE: we return a real, not an int, as with deterministic mode this
+// will not necessarily be an integer
```
johnlees, line 170: Might need to be careful with that, as without changes in the odin model could lead to a non-integer n at the next step.
Rounding every step (especially if dt decreases) leads to issues. The thing we found was best with the rnorm based simulation was to keep the real at each dt step, and then round at the data steps.
I suppose this is an odin/model issue rather than anything that can be changed here (I think rounding the result would remove any flexibility around this issue)

johnlees: Looking at the wikipedia page, floor/ceil of np is the median

inst/include/dust/distr/poisson.hpp
```diff
@@ -9,107 +9,131 @@ namespace distr {
...
+// NOTE: we return a real, not an int, as with deterministic mode this
+// will not necessarily be an integer
+__nv_exec_check_disable__
+template <typename real_t>
+HOSTDEVICE real_t rpois(rng_state_t<real_t>& rng_state,
+                        typename rng_state_t<real_t>::real_t lambda) {
+  real_t x = 0;
   if (lambda == 0) {
     // do nothing, but leave this branch in to help the GPU
+#ifndef __CUDA_ARCH__
```
johnlees, line 129: Why do we have these includes? Is deterministic + GPU disallowed?
I appreciate that it would likely be inefficient, but might be confusing behaviour to remain stochastic

inst/include/dust/dust.hpp
```diff
@@ -235,9 +241,14 @@ class Dust {
     }
   }
 
-  std::vector<real_t> simulate(const std::vector<size_t>& step_end) {
+  std::vector<real_t> simulate(const std::vector<size_t>& step_end,
+                               bool deterministic) {
```
johnlees, line 245:
```suggestion
                               const bool deterministic) {
```

inst/include/dust/dust.hpp
```diff
@@ -147,19 +147,22 @@ class Dust {
...
     }
     const auto r = std::minmax_element(step.begin(), step.end());
     if (*r.second > *r.first) {
-      run(*r.second);
+      run(*r.second, deterministic);
     }
   }
 
-  void run(const size_t step_end) {
+  void run(const size_t step_end, bool deterministic) {
```
johnlees, line 161:
```suggestion
  void run(const size_t step_end, const bool deterministic) {
```

inst/include/dust/dust.hpp
```diff
@@ -147,19 +147,22 @@ class Dust {
     }
   }
 
-  void set_step(const std::vector<size_t>& step) {
+  void set_step(const std::vector<size_t>& step, bool deterministic) {
```
johnlees, line 150:
```suggestion
  void set_step(const std::vector<size_t>& step, const bool deterministic) {
```

inst/include/dust/distr/poisson.hpp
johnlees: Ok I see the check in the interface below, so this makes sense now

inst/include/dust/xoshiro.hpp
```diff
@@ -140,29 +144,34 @@ inline HOST void xoshiro_long_jump(rng_state_t<T>& state) {
...
 
 template <>
 inline HOSTDEVICE float unif_rand(rng_state_t<float>& state) {
-  const uint64_t value = xoshiro_next(state);
 #ifdef __CUDA_ARCH__
+  const uint64_t value = xoshiro_next(state);
   float rand = (__fdiv_rn(__ull2float_rn(value), 18446744073709551616.0f));
 #else
+  if (state.deterministic) {
+    return 0.5;
```
johnlees, line 172:
```suggestion
    return 0.5f;
```

tests/testthat/test-example.R
```diff
@@ -702,3 +702,52 @@ test_that("steps must not be negative", {
...
+  state[2, ] <- state[2, ] + n
+  step <- 1:10
+
+  mod$set_state(state, step, deterministic = TRUE)
+
+  res <- mod$state()
+  f <- function(i) {
+    mod <- sir$new(list(), 0, 1, seed = NULL)
+    mod$set_state(state[, i], step[i])
+    mod$run(10, deterministic = TRUE)
```
johnlees, line 749: Does the SIR example round the result from rbinom?

johnlees (changes requested):

inst/include/dust/distr/binomial.hpp
richfitz: Hmm, thanks for thinking about this. We *do* want non-integer n here, and I'll have to think about the way we handle that round, most likely moving it into the rbinom call

johnlees: Also in the limit of small step size, for this to work do you need to take a real for the input `n` too right (to get the mean)? Maybe another function/template to deal with the deterministic case would be best

richfitz: problem with a template is we then only have compile time polymorphism - I'll have a little think about a decent way of hiding the complexity, but fundamentally `rbinom` will need to change from `(int, real) -> int` to `(real, real) -> real` for this to work, and the round will need to move inside the nondeterministic arm

tests/testthat/test-example.R
richfitz: no, so these non-integral numbers will persist. This is what Neil was looking for (and of course was impossible prior to changes in 0c4915c as otherwise rbinom was truncating `n`)

inst/include/dust/distr/binomial.hpp
```diff
@@ -166,34 +166,50 @@ inline HOSTDEVICE real_t btrs(rng_state_t<real_t>& rng_state, int n_int, real_t
   return draw;
 }
 
+template <typename real_t>
+HOSTDEVICE void rbinom_validate(int n, real_t p) {
```
johnlees, line 170: template the n type here instead of implicitly casting on L197 below?

johnlees: Or explicitly cast on L197

johnlees (approved):

inst/include/dust/distr/binomial.hpp
richfitz: Added a cast - can't be templated because of the %d in the sprintf expression

inst/include/dust/distr/binomial.hpp
```diff
@@ -166,34 +166,50 @@ inline HOSTDEVICE real_t btrs(rng_state_t<real_t>& rng_state, int n_int, real_t
...
+  rbinom_validate(n, p);
+  return n * p;
+}
+
 // NOTE: we return a real, not an int, as with deterministic mode this
 // will not necessarily be an integer
 template <typename real_t>
-HOSTDEVICE real_t rbinom(rng_state_t<real_t>& rng_state, int n,
-                         typename rng_state_t<real_t>::real_t p) {
+HOSTDEVICE real_t rbinom_stochastic(rng_state_t<real_t>& rng_state, int n,
```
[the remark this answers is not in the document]
richfitz: I prefer all the functions having the same return type here, practically it will make very little difference though

PR #253, opened 2021-08-10
Reorganise headers

### LINKED ISSUE #228: Consistent header names (hpp/cuh); GPU subfolder
I suggest we just rename everything as .hpp, the distinction feels arbitrary here as files always contain both C++ and CUDA, and remembering which suffix to use is annoying.

Also I think we can tidy the repository by creating some subfolders, a CUDA one at least

richfitz:
This PR moves the cuda-related headers to the inst/include/dust/cuda subdir, renames .cuh to .hpp and updates their headerguards (and includes).

Merge after #252, or merge this PR which includes it

Fixes #228

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: dust
 Title: Iterate Multiple Realisations of Stochastic Models
-Version: 0.9.12
+Version: 0.9.13
 Authors@R: c(person("Rich", "FitzJohn", role = c("aut", "cre"),
                     email = "rich.fitzjohn@gmail.com"),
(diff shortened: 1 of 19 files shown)


DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: dust
 Title: Iterate Multiple Realisations of Stochastic Models
-Version: 0.9.11
+Version: 0.9.12
```
johnlees, line 3:
```suggestion
Version: 0.9.13
```
As I merged #252

johnlees (approved):

PR #264, opened 2021-09-27
documentation update

richfitz:
Expand examples, and some other documentation issues

* add examples to all help files
* add return values to all help files
* fix links

Fixes #262

PR #267, opened 2021-09-27
Control compilation and cache

### LINKED ISSUE #204: Generate but don't compile, without package
This will be useful for the sircovid code

### LINKED ISSUE #248: Option to skip cache
Can't compile two models with different gpu flags without a false cache hit - perhaps just expose an option to skip this, and odin.dust can pass that along

richfitz:
This PR fixes two closely related issues

* allow skipping the cache in compilation #248, useful when working with odin.dust on gpu models
* allow creating code without compiling it #204, again useful with gpu models

Merge after #264, which this builds on

Fixes #248 
Fixes #204


johnlees (approved):
This all looks fine to me. Only thought was whether it would be useful to document `dust_generate` anywhere in the developer/GPU vignette(s)? I'll leave it up to you to decide, I can't remember if this fits neatly in to them or not

PR #274, opened 2021-09-29
Generalised reset/update state approach

### LINKED ISSUE #180: Tidy up reset methods
* set_pars: sets just pars
* reset: sets pars and time
* set_state: sets state and/or time
* set_step: just time

I think we could rationalise these effectively. Perhaps a single method?

### LINKED ISSUE #188: Remove some comments on file copy
Perhaps lines starting with `///` or `###` as it will be useful to leave notes:

* in dust.cpp changes to signatures need to be reflected into dust.hpp

### LINKED ISSUE #270: 'deterministic' should be a fixed property of class at init, not variable

richfitz:
This PR combines the three different state setting functions we had into one universal function:

* `reset` updated pars and time and set initial conditions
* `set_state` updated state and maybe time
* `set_pars` updated pars but did not set initial conditions

The new function does all this and more, with a simpler more consistent interface

I need to pull together a slightly more substantial PR for mcstate/sircovid/spimalot etc but it's not too bad really

* mcstate: https://github.com/mrc-ide/mcstate/pull/147
* sircovid: https://github.com/mrc-ide/sircovid/pull/349

All the "interesting" bits of the PR are in the inst/include files really, changes to the tests are extensive but mostly consist of swapping out calls fairly robotically

Fixes #180 
Fixes #188 (as I was adding some of these comments as I went)
Fixes #270 (as this covers much of the same ground and I can group disruptive changes this way)


R/dust_generator.R
```diff
@@ -250,6 +243,35 @@ dust_generator <- R6::R6Class(
     set_pars = function(pars) {
     },
 
+    ##' @description Update one or more components of the model state.
+    ##'   This method can be used to update any or all of `pars`, `state` and
+    ##'   `step`.  If `pars` is given and `state` is not, then by default we
+    ##'   will update the model internal state according to your model's
+    ##'   initial conditions - use `set_initial_state = FALSE` to prevent
+    ##'   this.
```
johnlees, line 251: I might expect the opposite behaviour as default, but I presume this way is how this is being used by sircovid etc

inst/include/dust/dust.hpp
```diff
@@ -145,22 +135,31 @@ class Dust {
...
+  // range [0, n-1]
+  void set_index(const std::vector<size_t>& index) {
+    index_ = index;
+    update_device_index();
+  }
+
+  void run(const size_t step_end) {
     refresh_host();
-    if (deterministic) {
+    if (deterministic_) {
```
johnlees, line 162: Now this is no longer dynamic, should `rng_.set_deterministic()` be moved to the constructor?

inst/include/dust/dust.hpp
```diff
@@ -241,12 +240,11 @@ class Dust {
...
   }
 
-  std::vector<real_t> simulate(const std::vector<size_t>& step_end,
-                               const bool deterministic) {
+  std::vector<real_t> simulate(const std::vector<size_t>& step_end) {
     const size_t n_time = step_end.size();
     std::vector<real_t> ret(n_particles() * n_state() * n_time);
 
-    if (deterministic) {
+    if (deterministic_) {
```
johnlees, line 247: Same here -> move to constructor

johnlees: (unless it does get changed while the object is being used?)

inst/include/dust/interface.hpp
```diff
@@ -100,15 +102,87 @@ void dust_set_index(SEXP ptr, cpp11::sexp r_index) {
...
+      info_list[i] = dust_info<T>(pars[i]);
+    }
+    obj->set_pars(pars, set_initial_state);
+    ret = info_list;
+  }
+  return ret;
+}
+
+template <typename T>
+cpp11::sexp dust_update_set(Dust<T> *obj, SEXP r_pars,
```
johnlees, line 130: or just `dust_update` or `dust_set`?

inst/include/dust/interface.hpp
```diff
@@ -100,15 +102,87 @@ void dust_set_index(SEXP ptr, cpp11::sexp r_index) {
...
+  if (state.size() > 0 && step.size() > 0) {
+    obj->reset_errors();
+  }
+
+  return ret;
+}
 
+// There are many components of state (not including rng state which
+// we treat separately), we set components always in the order (1)
+// pars, (2) state, (3) step
```
johnlees, line 161: Does this comment belong to the above function?

johnlees (changes requested):

inst/include/dust/dust.hpp
richfitz: You can't tell on the constructor if the model is going to be used on the device (we accept `NULL` device information)

R/dust_generator.R
richfitz: Indeed, that's the trick. We could change this to be if we have pars AND step NOT state, perhaps, which is the least surprising set and I think fully compatible with sircovid (this was just the old `reset` method)

inst/include/dust/dust.hpp
richfitz: oh sorry, it is there. I must be doing something daft here

richfitz: Right; we need the rng to be usable in a non-deterministic way for compare and theoretically for the filter, so this dynamism is required here, even though it's a bit surprising. The consistent way of doing this would be I think to check at set_data and prevent anyone adding data, which in turn prevents use of compare/filter

inst/include/dust/interface.hpp
richfitz: other direction I think, `dust_update_state_set` as this is just a bit of breaking apart what was a monster method

inst/include/dust/interface.hpp
richfitz: yeah, that's the top-level function - the others come first due to scoping rules

R/dust_generator.R
richfitz: Updated to be less enthusiastic about setting initial conditions

inst/include/dust/interface.hpp
```diff
@@ -100,15 +102,88 @@ void dust_set_index(SEXP ptr, cpp11::sexp r_index) {
...
+    ret = info_list;
+  }
+  return ret;
+}
+
+template <typename T>
+cpp11::sexp dust_update_state_set(Dust<T> *obj, SEXP r_pars,
+                                  const std::vector<typename T::real_t>& state,
+                                  const std::vector<size_t>& step,
+                                  bool set_initial_state) {
```
johnlees, line 133:
```suggestion
// There are many components of state (not including rng state which
// we treat separately), we set components always in the order (1)
// pars, (2) state, (3) step
template <typename T>
cpp11::sexp dust_update_state_set(Dust<T> *obj, SEXP r_pars,
                                  const std::vector<typename T::real_t>& state,
                                  const std::vector<size_t>& step,
                                  bool set_initial_state) {
```

richfitz: made this change but moved rather than added the comment

johnlees (approved):

PR #284, opened 2021-10-06
Use uint32 for float rng

### LINKED ISSUE #282: Use 32 bit rng for floats in dust_rng object

richfitz:
Part of the RNG epic, merges into develop (see #283)

* Use 32 bit generator for the float rng in `dust::dust_rng` for R
* Use the C api names (rbinom becomes binomial)

These cause quite a bit churn in the tests, but are simple in implementation. In particular the previous case where the 64 bit rng created floats was checked differently because the rng proceeded the same way as before, only the created float differed. Now we compare against the expectations

Fixes #282

### DIFF
--- a/R/rng.R
+++ b/R/rng.R
@@ -95,11 +95,6 @@ dust_rng <- R6::R6Class(
       private$n_generators
     },
 
-    ##' @description Indicates the floating point type
-    real_type = function() {
-      if (private$float) "float" else "double"
-    },
-
     ##' @description The jump function for the generator, equivalent to
     ##' 2^128 numbers drawn from the generator.
     jump = function() {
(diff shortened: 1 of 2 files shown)


R/rng.R
```diff
@@ -64,17 +68,25 @@ dust_rng <- R6::R6Class(
                                     private$float)
       private$n_generators <- n_generators
 
+      if (real_type == "float") {
+        size_int_bits <- 32L
+        name <- "xoshiro128starstar"
+      } else {
+        size_int_bits <- 64L
+        name <- "xoshiro128starstar"
```
johnlees, line 76: Just checking these are the right choices. I might have expected the second one to be 256 bits

tests/testthat/test-rng.R
```diff
@@ -450,22 +441,20 @@ test_that("special case", {
   m <- 1000000
   n <- 6
   p <- 0.449999988
-  yf <- dust_rng$new(1, 1, "float")$rbinom(m, n, p)
-  yd <- dust_rng$new(1, 1, "double")$rbinom(m, n, p)
```
johnlees, line 454: double is still checked somewhere else?

johnlees (approved):

R/rng.R
richfitz: indeed this is wrong

Issue #280, opened 2021-10-04
int_to_real needs proper implementation


richfitz:

johnlees:
Possibilities, based on these:
## simple 
consts can be found with `printf("%x")` which uses enough precision
```
#define 2POW32_INV (2.3283064e-10f)
#define 2POW32_INV_DOUBLE (2.3283064365386963e-10) 
#define 2POW53_INV_DOUBLE (1.1102230246251565e-16)
```
32 bit int -> float
```
return x * 2POW32_INV + (2POW32_INV/2.0f);
```
32 bit int -> double
```
return x * 2POW32_INV_DOUBLE + (2POW32_INV_DOUBLE/2.0);
```
64 bit int -> float (uses the top 32 bits)
```
unsigned int t;
t = (unsigned int)(x >> 32); 
return t * 2POW32_INV + (2POW32_INV/2.0f);
```
64 bit int -> double (uses 53 bits from the mantissa, as suggested in randomdouble.pdf)
```
return (x >> 11) * 2POW53_INV_DOUBLE + (2POW53_INV_DOUBLE/2.0);
```

## Doornik method (doubles)
```
#define M_RAN_INVM32 2.32830643653869628906e-010
#define M_RAN_INVM52 2.22044604925031308085e-016
```
32 bit int -> double
```
return (int)(x) * M_RAN_INVM32 + (0.5 + M_RAN_INVM32 / 2)
```
two 32-bit ints x1, x2 -> double
```
return (int)(x1) * M_RAN_INVM32 + (0.5 + M_RAN_INVM52 / 2) + (int)((x2) & 0x000FFFFF) * M_RAN_INVM52
```
I think the 64 bit versions or float versions should be straightforward from this and the above

## Campbell method
This is 64-bit rng (possibly multiple draws, but usually two) to double

HOSTDEVICE double random_real(rng_state_t& rng) {
	int exponent = -64;
	uint64_t significand;
	unsigned shift;
	double draw = 1.0;

#ifdef __CUDA_ARCH__
	while (significand = rng.next()) == 0)
#else
	while (__predict_false((significand = rng.next()) == 0)) {
#endif
		exponent -= 64;
#ifdef __CUDA_ARCH__
		if (exponent < -1074)
#else
		if (__predict_false(exponent < -1074))
#endif
			draw = 0;
			break;
	}

	if (draw != 0) {
#ifdef __CUDA_ARCH__
		shift = __clzll(significand);
#else
[...]

PR #285, opened 2021-10-06
I280 int to real

richfitz:
Use the "simple" method in #280 for conversion from int to real

I think this will be both fast and accurate enough for us here


inst/include/dust/random/utils.hpp
```diff
@@ -8,25 +8,40 @@ namespace random {
...
+double int_to_real(uint64_t x) {
+  return (x >> 11) * TWOPOW53_INV_DOUBLE + (TWOPOW53_INV_DOUBLE / 2.0);
+}
 
 template <>
 inline HOSTDEVICE
-double int_to_real(uint64_t value) {
-  return double(value) / double(utils::uint64_max());
+double int_to_real(uint32_t x) {
+  return (x >> 11) * TWOPOW32_INV_DOUBLE + (TWOPOW32_INV_DOUBLE / 2.0);
```
johnlees, line 31:
```suggestion
  return x * TWOPOW32_INV_DOUBLE + (TWOPOW32_INV_DOUBLE / 2.0);
```

PR #290, opened 2021-10-11
Remaining RNG changes: parallel R interface, standalone support, new types, docs

### LINKED ISSUE #279: Parallel rng in example rng object
This will require some thought as we currently do:

```
  for (size_t i = 0; i < (size_t)n; ++i) {
    y[i] = dust::random::uniform<real_t>(rng->state(i % n_generators),
                                         min[i], max[i]);
  }
```

this can't be portably parallelised unless we are looping over generators

```
#pragma omp parallel for schedule(static_) num_threads(num_threads_)
for (size_t i = 0; i < rng->size(); ++i) {
  ...
}
```

but now we struggle to match the interface we have before. One obvious solution would be to generate the required numbers on every generator?

### LINKED ISSUE #286: Helper for creating sensible default rng based on real type
something like

```cc
namespace dust {

template <typename T>
struct default_rng_type_helper;

template <>
struct default_rng_type_helper<double> {
  typedef dust::random::xoshiro256starstar type;
};

template <>
struct default_rng_type_helper<float> {
  typedef dust::random::xoshiro256starstar type;
};

template <typename T>
using default_rng_type = default_rng_type_helper<T>::type;
```

then in the model

```cc
class walk {
  typedef double real_type;
  typedef dust::default_rng_type<real_type> rng_state_type;
  ...
};
```

with better names throughout

richfitz:
This is hopefully the last PR into #283 before we look towards merging that PR.

It introduces a hodgepodge of changes I'm afraid:

* The R rng `dust::dust_rng` is now parallel and has a more useful approach to using multiple generators. This is quite disruptive, but a big improvement.
* I've added the xoshiro512 generators because why not
* I've added some docs to rng.Rmd about the different generators, reusing them from a package and reusing them in a standalone way
* There's a new helper type `dust::random::generator<real_type>` which will select a reasonable rng (either xoshiro256+ or xoshiro128+) **NOTE** this changes the default type from `**` to `++` which is a little faster

In the meantime I have bumped the version to 0.10.0

Fixes #279 
Fixes #286

### DIFF
--- a/.github/workflows/make-release.yaml
+++ b/.github/workflows/make-release.yaml
@@ -154,7 +154,7 @@ jobs:
           asset_name: ${{ env.PACKAGE_NAME }}_${{ env.PACKAGE_VERSION }}_oldrel.zip
           asset_content_type: application/zip
 
-      - name: Build random library
+      - name: Create archive of random library
         shell: bash
         run: |
           tar -zcvf dust-random.tar.gz -C inst/include dust/random
--- a/vignettes/rng.Rmd
+++ b/vignettes/rng.Rmd
@@ -186,7 +186,7 @@ The supported types are:
 * `xoshiro256starstar`, `xoshiro256plusplus`, `xoshiro256plus` (4 x 64 bit state)
 * `xoroshiro128starstar`, `xoroshiro128plusplus`, `xoroshiro128plus` (2 x 64 bit state)
 * `xoshiro128starstar`, `xoshiro128plusplus`, `xoshiro128plus` (4 x 32 bit state)
-* `xoshiro256starstar`, `xoshiro256plusplus`, `xoshiro256plus` (8 x 64 bit state; this is far more state space than typically needed)
+* `xoshiro512starstar`, `xoshiro512plusplus`, `xoshiro512plus` (8 x 64 bit state; this is far more state space than typically needed)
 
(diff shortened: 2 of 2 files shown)


.github/workflows/make-release.yaml
```diff
@@ -153,3 +153,18 @@ jobs:
           asset_path: pkg/pkg-windows-latest-oldrel/${{ env.PACKAGE_NAME }}_${{ env.PACKAGE_VERSION }}.zip
           asset_name: ${{ env.PACKAGE_NAME }}_${{ env.PACKAGE_VERSION }}_oldrel.zip
           asset_content_type: application/zip
+
+      - name: Build random library
```
johnlees, line 157:
```suggestion
      - name: Create archive of random library
```
Just as this is the code, not a built executable right? Might be a little clearer

vignettes/rng.Rmd
```diff
@@ -150,3 +154,117 @@ bench::mark(
...
+```
+
+which means that whatever real type you use, you want to use the Xoshiro256** generator.
+
+The supported types are:
+
+* `xoshiro256starstar`, `xoshiro256plusplus`, `xoshiro256plus` (4 x 64 bit state)
+* `xoroshiro128starstar`, `xoroshiro128plusplus`, `xoroshiro128plus` (2 x 64 bit state)
+* `xoshiro128starstar`, `xoshiro128plusplus`, `xoshiro128plus` (4 x 32 bit state)
+* `xoshiro256starstar`, `xoshiro256plusplus`, `xoshiro256plus` (8 x 64 bit state; this is far more state space than typically needed)
```
johnlees, line 189:
```suggestion
* `xoshiro512starstar`, `xoshiro512plusplus`, `xoshiro512plus` (8 x 64 bit state; this is far more state space than typically needed)
```

PR #302, opened 2021-11-01
Decouple GPU and CPU models

### LINKED ISSUE #154: Refactor god class
Possibly consider around same time as #145, we need to break up the Dust class as it's a bit out of control. Much of this could just be grouping together bits of data into structs rather than leaving everything free, but we might also break the object into a lower-level and higher-level class too

### LINKED ISSUE #254: reset with vector of times on the gpu will be slow!
Currently this will run on the cpu...

### LINKED ISSUE #292: Decouple device and non-device models
I think this is the trick to fixing #154:

* Device is selected on initialisation or not, and we disallow any switching between host/device during running
* Probably completely new class that is the gpu version

I don't know that this will break any existing code, though it will require regenerating everything so it should be done before the CRAN release.

richfitz:
Another total redesign of the way that gpu models are included, reflecting how we actually use these now.

* on compilation we might or might not support including gpu-like code. This is detected automatically or can be forced with the cpp11 pseudo-attribute `dust::has_gpu_support` (this supercedes the old `has_gpu_support` template as this is useful to know ahead of time).
* on initialisation we might point at a device or not; that instance will then run entirely on either gpu or cpu
* all the run/simulate/compare_data/filter functions lose their old `device` argument

This PR will represent a mid-point along a series of smaller cleanups, as the current gpu code is still a bit redundant.

Some thoughts on future cleanups that we might move into separate PRs

* longer term it would be nice to exclude all cuda code from being included if not used (e.g., in interface/dust.hpp do we really want to include cuda/filter.hpp?)
* rework `filter_state_type` definition in Dust and DustDevice? (hard to due disabled `=`/move constructor)
* device_info should not be the only source of `real_bits`
* redefine name of` __nv_exec_check_disable__`
* move all cuda predefines somewhere into random
* update vignette, which is now wrong (already has issue)
* split the test-gpu.R test file into support and running

Fixes #292 
Fixes #154 (or close enough anyway)
Fixes #254 (by preventing the problem for now)


inst/include/dust/interface/cuda.hpp
```diff
@@ -0,0 +1,73 @@
...
+    cpp11::sexp r_run_block_size = r_device_config_l["run_block_size"];
+    if (r_run_block_size != R_NilValue) {
+      int run_block_size_int = cpp11::as_cpp<int>(r_run_block_size);
+      if (run_block_size_int < 0) {
+        cpp11::stop("'run_block_size' must be positive (but was %d)",
+                    run_block_size_int);
+      }
+      if (run_block_size_int % 32 != 0) {
+        cpp11::stop("'run_block_size' must be a multiple of 32 (but was %d)",
+                    run_block_size_int);
```
johnlees, line 50: Sometimes setting block size and block count = 1 is useful for debugging so everything is serial

inst/include/dust/interface/dust.hpp
```diff
@@ -193,10 +213,14 @@ SEXP dust_update_state(SEXP ptr, SEXP r_pars, SEXP r_state, SEXP r_step,
     // best thing to take for those? Possibly require an array to
     // disambiguate?
     step = dust::interface::validate_size(r_step, "step");
-    if (!(step.size() == 1 || step.size() == obj->n_particles())) {
+    const size_t len = step.size();
+    if (!(len == 1 || len == obj->n_particles())) {
       cpp11::stop("Expected 'step' to be scalar or length %d",
                   obj->n_particles());
     }
+    if (!std::is_same<T, Dust<typename T::model_type>>::value && len != 1) {
```
johnlees, line 221: yikes

inst/template/dust_methods.cpp
```diff
@@ -0,0 +1,77 @@
+/// IMPORTANT; changes here must be reflected in inst/template/dust_methods.hpp
```
johnlees, line 1: Can you update these files in the developer notes, I would probably forget to look here

johnlees (changes requested):

inst/include/dust/cuda/dust_device.hpp
```diff
@@ -0,0 +1,797 @@
...
+                     device_state_.rng.data(),
+                     use_shared_int,
+                     use_shared_real);
+#endif
+  }
+
+private:
+  // delete move and copy to avoid accidentally using them
+  DustDevice ( const DustDevice & ) = delete;
+  DustDevice ( DustDevice && ) = delete;
```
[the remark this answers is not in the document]
richfitz: The profiler destructor does not though, and that causes issues as on move profiling stops!

inst/include/dust/interface/dust.hpp
richfitz: Why yikes here? this is less hairy than some of the template magic we had before :upside_down_face:

inst/include/dust/cuda/dust_device.hpp
```diff
@@ -0,0 +1,797 @@
...
+#endif
+    for (size_t i = 0; i < np; ++i) {
+      dust::utils::destride_copy(end_state + i * index_size, y_selected, i,
+                                  np);
+    }
+  }
+
+  // TODO: we should really do this via a kernel I think? Currently we
+  // grab the whole state back from the device to the host, then
+  // filter through it.
```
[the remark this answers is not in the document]
richfitz: yeah, this is all stuff for later (and is the same as previous)

inst/include/dust/cuda/dust_device.hpp
```diff
@@ -0,0 +1,797 @@
...
+      dust::cuda::device_shared_copy<T>(pars[i].shared, dest_int, dest_real);
+    }
+    device_state_.shared_int.set_array(shared_int);
+    device_state_.shared_real.set_array(shared_real);
+  }
+
+  // TODO: This update function is wildly inefficient; we should
+  // probably support things like "copy one state to all the particles
+  // of that parameter index", possibly as a kernel.
+  void set_state_from_pars(const std::vector<pars_type>& pars) {
```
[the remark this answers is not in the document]
richfitz: It can be yes, but I am not sure we cope correctly with that yet and we do not test it anywhere - I think this is wrong in CPU code too https://github.com/mrc-ide/dust/issues/310 (added to comment too)

inst/include/dust/interface/cuda.hpp
richfitz: This is unchanged from before - I'm inclined not to change it here atm as for debugging we can hack it in

inst/template/dust_methods.cpp
richfitz: done

inst/include/dust/cuda/dust_device.hpp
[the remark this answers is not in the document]
richfitz: https://github.com/mrc-ide/dust/issues/311 posted for this now

johnlees (approved):

PR #312, opened 2021-11-03
I307 multinomial

### LINKED ISSUE #307: Add support for multinomial
* make sure it works with non-normalised prob
* complex handling of args for the R interface, unfortunately

richfitz:
This PR adds support for the multinomial distribution

cc: @annecori 

Fixes #307


R/rng.R
```diff
@@ -50,6 +50,27 @@
...
+##'   * If a matrix with `n` columns then vary over each draw (the
+##'     `i`th draw using vector `prob[, i]` but shared across all
+##'     generators. There are `nrow(prob)` possible outcomes.
+##'
+##'   * If a 3d array is provided with 1 column and `n_generators`
+##'     "layers" (the third dimension) then we use then we use different
+##'     parameters for each generator, but the same parameter for each
+##'     draw.
+##'
+##'   * If a 3d array is providec with `n` columns and `n_generators`
```
johnlees, line 69:
```suggestion
##'   * If a 3d array is provided with `n` columns and `n_generators`
```

inst/include/dust/random/multinomial.hpp
```diff
@@ -0,0 +1,74 @@
...
+
+#include <cmath>
+#include <dust/random/numeric.hpp>
+
+namespace dust {
+namespace random {
+
+// This is written assuming that 'prob' and 'ret' are arbitrary
+// containers; they could be pointers, vectors or anything else that
+// supports random access.  We cold also do this with iterators but
```
johnlees, line 12:
```suggestion
// supports random access.  We could also do this with iterators but
```

tests/testthat/test-rng.R
```diff
@@ -515,6 +515,174 @@ test_that("exponential random numbers from floats have correct distribution", {
...
+test_that("multinomial algorithm is correct", {
+  set.seed(1)
+  prob <- runif(10)
+  prob <- prob / sum(prob)
+  size <- 20
+  n <- 5
+
+  res <- dust_rng$new(1, seed = 1L)$multinomial(n, size, prob)
+
+  ## Separate implementation of the core algorithm:
```
johnlees, line 527: Also compare with `rmultinom()` results?

johnlees: I can't remember how the other distributions are tested, whether they're still compared to the R implementations, or if it's all done via mean and stddev now

johnlees (approved):
Any plans for a CUDA version of this? An iterator version of the core function would work for both, but I could see that allocating prob and return vectors might be more involved (especially if their length ever changes)

tests/testthat/test-rng.R
richfitz: all against expectations now

richfitz:
I've made an issue for GPU support: https://github.com/mrc-ide/dust/issues/315 - practically we can't use this from odin.dust atm so I don't think this is urgent, and it _should_ be really easy with the iterator based version

Issue #271, opened 2021-09-28
Add simple python bindings to the dust class


johnlees:
These don't necessarily have to have all the same functionality as the R bindings, but it would be useful to use the dust class directly from python (e.g. for use with the ELFI package)

PR #321, opened 2021-11-05
Rationalise names used for gpu things

### LINKED ISSUE #298: Rename 'device' to 'gpu' everwhere
We use 'cuda', 'device' and 'gpu' interchangeably; all should move to gpu so that it is easier to explain

### LINKED ISSUE #301: add better accessor for real info
Currently it's in device_info, which is silly. Would be nice to return name and/or size elsewhere

### LINKED ISSUE #304: Inconsistent style
We use Dust, DustDevice and Particle but all other names are lower case

### LINKED ISSUE #317: Consistent namespaces
* Possibly at same time as #298 as it overlaps, make sure we're consistent about use of `interface` and `r` namespace. 
* `cuda` becomes `gpu` possibly.
* `dust::cuda::device_info` moves into an interface namespace

richfitz:
This PR rationalises names used for gpu things:

* most uses of device/cuda become gpu, especially when facing the user; this affects files, namespace and functions!
* exceptions include "device_id", actual device storage (an implementation detail really) and things to do with the compilation toolchain
* At the same time (to minimise the number of disruptive changes) the `interface` namespace becomes `r`, which will make #271 much easier
* New check to make sure that header guards are correct

I've tried to avoid folding any features in here, but couldn't help adding a fix for #301 because I was interacting with code that is very nearby.

Fixes #317
Fixes #304 
Fixes #301 
Fixes #298


NEWS.md
```diff
@@ -1,3 +1,8 @@
+# dust 0.11.5
+
+* Major header reorganisation, the `dust::cuda` namespace is now `dust::gpu`, most user-facing uses of "cuda" and "device" now replaced with "gpu (#298, #317)
```
johnlees, line 3:
```suggestion
* Major header reorganisation, the `dust::cuda` namespace is now `dust::gpu`, most user-facing uses of "cuda" and "device" now replaced with "gpu" (#298, #317)
```

inst/include/dust/random/binomial.hpp
```diff
@@ -32,7 +33,9 @@ HOSTDEVICE real_type fast_pow(real_type x, int n) {
 
 __nv_exec_check_disable__
 template <typename real_type>
-real_type HOSTDEVICE binomial_inversion_calc(real_type u, int n, real_type p) {
+__host__ __device__
+real_type binomial_inversion_calc(real_type u, int n,
+                                                      real_type p) {
```
johnlees, line 38: has something gone wrong with the formatting here?

inst/include/dust/random/cuda_compatibility.hpp
```diff
@@ -12,29 +12,21 @@
...
-// Compiling under gcc/clang
-#define HOST
-#define HOSTDEVICE
-#define DEVICE
-#define KERNEL
-#define ALIGN(n)
 #define __nv_exec_check_disable__
 
+// Additional stubs used to shadow the now-unneeded directives for
+// nvcc
```
johnlees, line 25:
```suggestion
// Additional stubs used to shadow the now-unneeded directives for
// nvcc, which would cause gcc/clang to raise an error
```

johnlees (approved):

Issue #325, opened 2021-11-08
Start on rng algorithms doc


richfitz:
Writing the ziggurat algorithm was unpleasant, add a vignette that explains how it is derived and keep track of any optimisations we add as they mostly reduce readability

Issue #324, opened 2021-11-08
Use integers for cutoff comparison


richfitz:
See Doornik and the wikipedia page; it's apparently faster to use a integer table of `x[i] / x[i - 1]` then compare the rng integer value against that. Requires that we keep two tables, one for uint32 and uint64 of course, and some benchmarking and explanation would be nice

richfitz:
This is going to be really hard as we can't easily get two versions of the coefficients for different integer types given the big fight in #327/#328; I still think it could be worth doing if faster but don't believe it will be _that_ much faster...

Issue #323, opened 2021-11-08
rng configuration object


richfitz:
Now that we have multiple algorithms possible, some sort of rng state object that includes:

* `real_type`
* `rng_state_type`
* normal algorithm
* and the state

Then make overloads for things like 

```
random_normal<real_type, algorithm, rng_state>(rng_state_type)
```

so that we can just write

```
random_normal(rng_state_type_wrapped)
```

something like

```
template <typename T>
typename T::real_type random_normal(T& state) {
  return random_normal<typename T::real_type, T::normal>(state.state);
}
```

See also #303 which is the same reallly

PR #326, opened 2021-11-08
Faster normal draws with the ziggurat algorithm

### LINKED ISSUE #308: Add alternative normal distribution algorithm
Our current one, BM, is fairly slow and we could do better on all platforms with at least polar, possibly better with monty python. Ziggurat would be the one to to implement with the best performance on CPUs, apparently worse on GPU due to latency.

To make this easy we also need a way of selecting the normal algorithm as part of the rng spec, that's less straightforward unfortunately.

richfitz:
This PR implements the ziggurat algorithm for normally distributed numbers. 

There are some follow-on bits of work to do that I am avoiding in this PR because they're more likely to be disruptive and this is large enough atm

* #323
* #324
* #325

The current version does compile on a GPU, but runs fairly slowly due to the doubles

Fixes #308


inst/include/dust/random/normal.hpp
```diff
@@ -5,44 +5,46 @@
...
-  do {
-    u1 = random_real<real_type>(rng_state);
-    u2 = random_real<real_type>(rng_state);
-  } while (u1 <= epsilon);
-
-  SYNCWARP
-  return std::sqrt(-2 * std::log(u1)) * std::cos(two_pi * u2);
-}
-
+real_type random_normal(rng_state_type& rng_state) {
```
johnlees, line 23: Is the algorithm chosen at run time or when compiling the object? Wondering whether we could just template rather than using this function

inst/include/dust/random/normal_ziggurat.hpp
```diff
@@ -0,0 +1,75 @@
...
+      break;
+    }
+  } while (true);
+  return ret;
+}
+}
+
+// TODO: this will not work efficiently for float types because we
+// don't have float tables for 'x' and 'y'; getting them is not easy
+// without requiring c++14 either. The lower loop using 'x' could
```
johnlees, line 35: How come we are able to manage this with binomial draws?

inst/include/dust/random/normal_ziggurat.hpp
```diff
@@ -0,0 +1,75 @@
...
+// rewritten easily as a function though taking 'u1' and so allowing
+// full template specialisation. However by most accounts this
+// performs poorly onn a GPU to latency so it might be ok.
+__nv_exec_check_disable__
+template <typename real_type, typename rng_state_type>
+__host__ __device__
+real_type random_normal_ziggurat(rng_state_type& rng_state) {
+  using ziggurat::x;
+  using ziggurat::y;
+  constexpr size_t n = 256;
```
johnlees, line 45: Comment to explain this choice and whether it is tuneable?

inst/include/dust/random/normal_ziggurat.hpp
```diff
@@ -0,0 +1,75 @@
...
+    if (i == 0) {
+      ret = normal_ziggurat_tail<real_type>(rng_state, r, u0 < 0);
+      break;
+    }
+    const auto z = u0 * x[i];
+    const auto f0 = std::exp(-0.5 * (x[i] * x[i] - z * z));
+    const auto f1 = std::exp(-0.5 * (x[i + 1] * x[i + 1] - z * z));
+    const auto u1 = random_real<real_type>(rng_state);
+    if (f1 + u1 * (f0 - f1) < 1.0) {
+      ret = z;
```
johnlees, line 65: should this `break`?

scripts/update_ziggurat_tables
```diff
@@ -0,0 +1,98 @@
...
+  n <- 256
+  dat <- zig_constants(n)
+  r <- dat$r
+  v <- dat$v
+  x <- c(v / f(r), intervals(n, r, v))
+  y <- x[-1] / x[seq_len(n)]
+
+  format <- function(z) {
+    s <- vapply(z, deparse, "", control = "digits17")
+    paste(
```
johnlees, line 14: Do we know that this is enough precision? (is it based on being < 2.2e-16?)

scripts/update_ziggurat_tables
```diff
@@ -0,0 +1,98 @@
...
+    x * (f(0) - f(x)) - v
+  }
+  r <- uniroot2(g, c(1.4, 4), tol = tolerance)$root
+  v <- r * f(r) + f_int(r)
+  list(n = n, r = r, v = v)
+}
+
+
+## Helper for root polishing
+uniroot2 <- function(f, bounds, ..., scal = 10) {
```
johnlees, line 77: Does tolerance need scaling?

scripts/update_ziggurat_tables
```diff
@@ -0,0 +1,98 @@
...
+intervals <- function(n, r, v) {
+  x <- numeric(n)
+  for (i in seq_len(n - 1)) {
+    x[i] <- if (i == 1) r else f_inv(f(x[i - 1]) + v / x[i - 1])
+  }
+  x
+}
+
+
+zig_constants <- function(n, tolerance = 1e-10) {
```
johnlees, line 48: Enough precision here too?

scripts/update_ziggurat_tables
```diff
@@ -0,0 +1,98 @@
...
+  ## As for intervals but with more robustness to being out of bounds
+  intervals <- function(n, r, v) {
```
johnlees, line 50: I would consider adding a reference here

johnlees (changes requested):

inst/include/dust/random/normal_ziggurat.hpp
richfitz: The binomial algorithm doesn't need to use the random numbers; the tables are only used in `stirling_approx_tail`, so we have fully specialised templates on `real_type` (and some ugly names `k_tail_values_d` and `stirling_approx_tail_f`). With C++14 we can make these template variables which removes the naming problem but because the algorithm here needs to use one or two random numbers along side the constants we'd end up with trying to make a partially specialised template (providing real_type but leaving rng_state_type open).

There are some solutions, but I'd like to try and implement these separately and compare timings to make sure that we don't end up paying too much on the CPU case

inst/include/dust/random/normal_ziggurat.hpp
richfitz: My first shot at this was fully tunable using `std::array<real_type, int>` for all sorts  of useful `n`! But it was a pain to work with and we can't have `std::array` on the GPU...

inst/include/dust/random/normal_ziggurat.hpp
richfitz: ugh, yes

scripts/update_ziggurat_tables
richfitz: digits17 is the full precision of the underlying number

scripts/update_ziggurat_tables
richfitz: Tolerance comes through in the dots here

scripts/update_ziggurat_tables
richfitz: I think so - this is 2 orders more than we get than by default. To do this "properly" we'd want to use long doubles really. I've seen implementations with these numbers only ok to 5 digits though, and ours is more accurate than Doornik (only differs in the ~6th place I think)

inst/include/dust/random/normal.hpp
richfitz: that is a template - am I missing something?

richfitz: After discussion: it *feels* like one could overload this as we know the algorithm is known at compile time but C++ doesn't let us do that

inst/include/dust/random/normal_ziggurat.hpp
richfitz: I've added a note indicating where this comes from and why now

johnlees (approved):

PR #328, opened 2021-11-08
Improvements to ziggurat GPU/floating point performance

richfitz:
I was going to do this last but, it was actually easiest:

* depend on C++14 now
* use a [variable template](https://en.cppreference.com/w/cpp/language/variable_template) for the coefficients
* updates the box muller gpu version to use the __cosf intrinsic where possible


richfitz:
Note that this fails because we need to declare the variable inline (see https://stackoverflow.com/questions/64610024/duplicate-symbols-with-global-template-variable-using-clang) but that's not available until C++17

PR #330, opened 2021-11-10
Normal docs and tweaks

richfitz:
Documentation-driven development: 

* New polar algorithm for normally distributed random numbers; faster than Box-Muller but slower than Ziggurat
* Tweaked algorithm for Ziggurat so now ~10% faster by avoiding drawing a second random number when working with 64 bit integers
* Added a new vignette describing the internals of the normal sampling algorithms

Fixes #325

PR #333, opened 2021-11-11
Helpers for using rngs from other packages

### LINKED ISSUE #329: Helpers for standalone rng state manipulation

richfitz:
This PR implements helpers that I needed while helping Will with https://github.com/mrc-ide/viralload; if someone wants to use the dust RNGs from their package then they need a way of keeping the state alive easily. Here we add a little `dust_rng_pointer` class that does that, and which can be used with any of the 12 underlying generator types, and which can be safely unpacked in user C++ code, at which point they can use the usual dust functions.

There's some interesting bits in here:

* The pointers know what algorithm they are initialised with at runtime and can validate this at compile time even though that happens first.
* One use I expect is that users will feed pointers generated this way into things like `parallel::parLapply` which will serialise the object, setting the pointer to `NULL`. To help with this I'm keeping a rarely synced copy of the state on the object. This will work with the most common use case which is to create an object then send it elsewhere. In the case where the object has ever been used this will error as the state will be out of date.

I've added a small vignette that explains how to use this for the single-process case. There's more work to get the multi-process case working that will build on this, but gets closer to #297 and I think will be best dealt with separately because this is already fairly large. Open to suggestions for good distributed problems on that issue for the docs, otherwise I'll probably do simple mcmc.

~This PR builds on #330 and should be considered after that PR. For differences in the meantime see https://github.com/mrc-ide/dust/compare/i325-normal-docs...i329-rng-helpers~

Fixes #329


R/rng_pointer.R
```diff
@@ -0,0 +1,70 @@
+##' @title Create pointer to random number generator
+##'
+##' For external use, vignette coming soon.
```
johnlees, line 3: Vignette in this PR?

inst/include/dust/r/random.hpp
```diff
@@ -40,6 +53,113 @@ std::vector<typename rng_state_type::int_type> as_rng_seed(cpp11::sexp r_seed) {
...
+  auto *rng = new prng<rng_state_type>(n_streams, seed);
+  auto r_ptr = cpp11::external_pointer<prng<rng_state_type>>(rng);
+  auto r_state = rng_state_vector(rng);
+  return cpp11::writable::list({r_ptr, r_state});
+}
+
+// Start with the assumption that we'll pass in the R6 object, might
+// write a simpler version later.
+template <typename rng_state_type>
+prng<rng_state_type>* rng_pointer_get(cpp11::environment obj,
```
johnlees, line 112: Possible useful reference for #271. If we wanted to add this and support saving and loading, see the serialisation examples:
https://pybind11.readthedocs.io/en/stable/advanced/classes.html#pickling-support
https://github.com/pybind/pybind11/blob/master/tests/test_pickling.cpp

vignettes/rng.Rmd
```diff
@@ -321,6 +321,8 @@ plain_output(readLines(file.path(path_pkg, "NAMESPACE")))
 
 Finally, run `cpp11::cpp_register()` before compiling your package so that the relevant interfaces are created (`R/cpp11.R` and `cpp11/cpp11.cpp`).  A similar process would likely work with Rcpp without any dependency on cpp11.
 
+More interesting use with persistent streams is described in `vignette("rng_package.Rmd")`
```
johnlees, line 324: Not sure how clear 'persistent streams' is on its own. I think the use in the vignette below might be described (based on similar FFI packages) as 'using the dust RNG from C++', whereas another (this?) vignette might be 'using the dust RNG from R'

vignettes/rng_package.Rmd
```diff
@@ -0,0 +1,301 @@
...
+  return tot / static_cast<double>(n) * 4.0;
+}
+```
+
+This snippet looks much the same as above:
+
+* We've added `[[cpp::linking_to(dust)]]` and included the dust random interface (`dust/r/random.hpp`)
+* The first line of the function safely creates a pointer to the random state data. The template argument here (`<dust::random::xoshiro256plus_state>`) refers to the rng algorithm and matches `rng$algorithm`
+* The second line extracts a reference to the first (C++ indexing starting at 0) random number stream - this pair of lines is roughly equivalent to `GetRNGstate()` except that that the random numbers do not come from some global source
+* After that the listing proceeds as before proceeds as before, except there is no equivalent to `PutRNGstate()`
```
johnlees, line 127: Maybe explain again why this isn't needed (no global state to update back to)

johnlees (changes requested):
Maybe also add to the vignette explaining the use of serialisation and sync?

inst/include/dust/r/random.hpp
richfitz: The R serialisation has no hooking functionality, it's super annoying

johnlees (approved):
This is great, thanks

Issue #310, opened 2021-11-03
Check particle initialisation


richfitz:
* stochastic initial conditions are not applied on GPU, probably are on CPU
* time-dependent initial conditions are definitely not applied correctly on GPU, probably not on CPU

richfitz:
The stochastic initial condition part can be fixed - if we remove the call to `initial()` from the constructor in `particle` then we can move that elsewhere in the code

richfitz:
```
      particles_.reserve(n_particles);
      for (size_t i = 0; i < n_particles; ++i) {
        particles_.push_back(p);
      }
      shared_ = {pars.shared};
    } else {
#ifdef _OPENMP
      #pragma omp parallel for schedule(static) num_threads(n_threads_)
#endif
      for (size_t i = 0; i < n_particles; ++i) {
        particles_[i].set_pars(p, set_state);
      }
```

this is not ideal though as we're creating a particle in series but _not_ updating initial state, then later we *do* set the pars stochastically but don't use the time correctly

richfitz:
This would be easier too if we disallowed the vector of times

PR #344, opened 2021-11-16
I300 checks

### LINKED ISSUE #299: Standardise on typedef vs using

### LINKED ISSUE #300: add test that name and class can differ in template
Noticed that this:

https://github.com/mrc-ide/dust/blob/cae1b47e79c53d061c88c8bda25633adf398411a/inst/template/dust.cpp#L76

uses `{{name}}` not `{{class}}` which should fail if these do not have the same value

### LINKED ISSUE #305: Disallow changing number of particles in dust initialise
In dust.hpp:initialise, we have logic that _looks_ like it allows changing of particle sizes. I think some of it is vestigial though, and what is really being done here is the initial construction of particles. Documentation, removing the clear(), and making the test

```
if (particles_.empty()) {
 // initial setup
} else {
  // update
}
```

would be preferable

richfitz:
Continues tidying things up by fixing three small things:

* test that dust name and class can differ
* simpler code around particle initialisation 
* switch from typedef to using everywhere

The last one is the biggest but most boring.

I had thought about starting to tackle #310 but that's genuinely a bit broken at the moment.

Fixes #300
Fixes #305 
Fixes #299


extra/harness.cpp
```diff
@@ -4,16 +4,16 @@
...
+using int_type = uint32_t;
 constexpr size_t data_size = 4;
 #elif defined(XOROSHIRO128)
-typedef uint64_t int_type;
+using int_type = uint64_t;
 constexpr size_t data_size = 2;
-#elif defined(XOSHIRO512)
-typedef uint64_t int_type;
+#elif defined(XOSHIRO512
+)using int_type = uint64_t;
```
johnlees, line 16: what's going on with the bracket here?

inst/include/dust/dust_cpu.hpp
```diff
@@ -403,21 +403,20 @@ class dust_cpu {
       throw std::invalid_argument(msg.str());
     }
 
-    if (particles_.size() == n_particles) {
+    if (particles_.empty()) {
+      particles_.reserve(n_particles);
+      for (size_t i = 0; i < n_particles; ++i) {
+        particles_.push_back(p);
+      }
```
johnlees, line 410: Is this method or `.resize(n)` followed by `vec[i] = init` faster? Are they the same? I've often wondered but don't know the answer

inst/include/dust/dust_cpu.hpp
```diff
@@ -437,10 +436,16 @@ class dust_cpu {
       }
       n = p.back().size(); // ensures all particles have same size
     }
-    // TODO: I think that we should disalow this bit of logic and
-    // number of particles should be surely fixed at initial creation?
-    // See #305
-    if (particles_.size() == n_particles_total_) {
+
+    if (particles_.empty()) {
```
johnlees, line 440: Does this new initialisation change things on the GPU at all?

inst/include/dust/r/dust.hpp
```diff
@@ -470,7 +470,7 @@ cpp11::writable::doubles run_filter(T * obj,
 template <typename T, typename std::enable_if<!std::is_same<dust::no_data, typename T::data_type>::value, int>::type = 0>
 cpp11::sexp dust_filter(SEXP ptr, bool save_trajectories,
                         cpp11::sexp r_step_snapshot) {
-  // typedef typename T::real_type real_type;
+  // using real_type = typename T::real_type;
```
johnlees, line 473: Do we need this line/comment?

johnlees (changes requested):
Out of interest, what is the advantage to `using = ` over `typedef`?

extra/harness.cpp
richfitz: oops

inst/include/dust/dust_cpu.hpp
richfitz: resize relies on the thing being stored being default constructible - it will be constructed to that length.

As to faster:

* if resize is possible we have the cost of default-constructing the particle, followed by benefit of subsequently being able to construct them in parallel
* reserve/push wastes no time on default construction but can't be done in parallel

inst/include/dust/dust_cpu.hpp
richfitz: We never hold a vector on the gpu, and the current checks are better (they were what revealed this issue). However, they have more complex problems in #310

inst/include/dust/r/dust.hpp
richfitz: Almost certainly not, will get rid of that

richfitz:
> Out of interest, what is the advantage to using = over typedef?

The primary benefit is consistency. There are places where they are equivalent (all these uses) and there are places where using allows fancier aliasing like this:

```c++
template <typename T>
using generator = typename default_rng_helper<T>::type;
```

from dust/random/random.hpp

But once I started using both I never knew which I was using in different situations, and this removes me having to think any more than necessary

johnlees (approved):

PR #349, opened 2021-11-17
multinomial gpu

richfitz:
Enable multinomial on the gpu

See https://github.com/mrc-ide/dust-random-bench/pull/1 for use

Fixes #349

### DIFF
--- a/inst/include/dust/random/cuda_compatibility.hpp
+++ b/inst/include/dust/random/cuda_compatibility.hpp
@@ -39,4 +39,23 @@
 #define SYNCWARP
 #endif
 
+namespace dust {
+namespace utils {
+
+// We cannot throw errors in GPU code, we can only send a trap signal,
+// which is unrecoverable.
+__host__ __device__
+inline void fatal_error(const char *message) {
+#ifdef __CUDA_ARCH__
+  printf(message);
+  printf("\n");
+  __trap();
+#else
+  throw std::runtime_error(message);
+#endif
+}
+
+}
+}
+
 #endif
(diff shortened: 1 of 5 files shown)


inst/include/dust/random/cuda_compatibility.hpp
```diff
@@ -39,4 +41,23 @@
 #define SYNCWARP
 #endif
 
+namespace dust {
+namespace utils {
+
+// We cannot throw errors in GPU code, we can only send a trap signal,
+// which is unrecoverable.
+__host__ __device__
+inline void fatal_error(const char *message) {
```
johnlees, line 50: this might benefit from `__forceinline__`, but maybe the compiler works that out anyway

johnlees: Actually thinking about how it's called, probably doesn't matter

inst/include/dust/random/multinomial.hpp
```diff
@@ -26,23 +26,24 @@ namespace random {
 // with use from odin code.
 template <typename real_type, typename rng_state_type,
           typename T, typename U>
+__host__ __device__
```
johnlees, line 29: Do `prob` and `ret` need any special considerations? I suppose that they be benefit from interleaving, but maybe that could just be left as a comment for now.

johnlees (approved):

inst/include/dust/random/multinomial.hpp
richfitz: I think the way it currently is written, prob needs work as it currently requires a `real_type*`, but `ret` will automatically cope with an interleaved type. There's a linked pr that you might peek at where I think it's more obvious

richfitz: Ah, no this is just `const T&` for prob and `U&` for prob an ret respectively. So that means it should do the right thing.  Here's the other PR : https://github.com/mrc-ide/dust-random-bench/pull/1 we should have a look there (either in this PR or another) to wire this up properly.

inst/include/dust/random/cuda_compatibility.hpp
richfitz: The `inline` there is only really necessary because otherwise it causes problems at linkage :)

PR #350, opened 2021-11-18
Remove dust_rng_state_long_jump

### LINKED ISSUE #343: Remove dust::dust_rng_state_long_jump
dust::dust_rng_state_long_jump is not great, and was introduced to support smc2 in mcstate. We use a different approach for make_seeds in pmcmc there too.

Both approaches used are problematic if the user uses anything over than the xoshiro256 generators because the jump coefficients used are just those ones!

richfitz:
Obsoleted by the new distributed random state functionality, and working in mcstate

Merge after https://github.com/mrc-ide/mcstate/pull/157
Fixes #343


johnlees (approved):
Maybe helpful to add comment to NEWS.md as there's an API change? Although, I doubt anyone else is using the function, I've found it useful to check in changelogs when something has disappeared/stopped working

Issue #275, opened 2021-09-29
Versioned API?


richfitz:
If we want to make changes to the API when the package is on CRAN it would be painful - but if we version the headers based on API, for cran releases, we could help with that. Alternatively we could support vendoring of course

richfitz:
cpp_vendor seems like a good place to start...

PR #352, opened 2021-11-19
Versioning

### LINKED ISSUE #335: standalone library needs version information
otherwise we can't depend on things. Can generate this when we eject the lib, or could require it to match and enforce it in the version check (requiring a certain amount of churn in all future PRs)

richfitz:
This PR tidies up the versioning a bit:

* adds some constants to the random library (will be available everywhere), automatically generated by a script
* adds a check to our version check script that changes to the headers require updating the header version
* documents this in our developing docs

I was going to start on #275 and enable vendoring but that does look a bit more complicated still so will hold off on that

Fixes #335


johnlees (approved):
Nice approach.
Noting that when we do the python bindings this makes it easy to add the version there too with something like:
```
m.attr("__version__") = DUST_VERSION_STRING;
```

PR #355, opened 2021-11-23
I354 cpp docs

### LINKED ISSUE #354: Document C++ interface properly

richfitz:
This PR starts the process of documenting the C++ interface, which was getting increasingly awkward to do from vignettes.

* Basic doxygen file which outputs xml
* Sphinx + breathe to assemble that into something reasonable
* Splice that into the pkgdown docs

There's quite an exciting system for creating automatic examples, but I think it works pretty well. It's possible that either sphinx or doxygen supports this more natively but it's not obvious to me.

The main outstanding pain now is that it would be (much) nicer if we could sort out getting sphinx to inherit the pkgdown theme. That might not be that hard to do given they're both bootstrap based, but I'm not giving it a try yet.

Lots of points of improvements in the docs I expect, but I believe I've captured the key bits.  We'll need to add another round here for the GPU interface probably, and I was thinking that the main dust objects would benefit from one too before starting on any python interface (#271). However, that can wait too.

Once this is up I will describe it in a blog post. Much of this approach came from [this blog post](https://devblogs.microsoft.com/cppblog/clear-functional-c-documentation-with-sphinx-breathe-doxygen-cmake/).

Other things that snuck in while being documented:

* Our DUST_VERSION_STRING was incorrectly not a string
* The `import_state` method in `prng` was unnecessarily general and I've tidied that up

~The nominally excessive size is due to the crazy default Doxyfile - I had a quick go at hacking it down but broke things on the first attempt. I'll probably have another go tomorrow but that should not hold anything else up~ Another attempt and we've got a sensible size Doxyfile now

Fixes #354


inst/include/dust/random/generator.hpp
```diff
@@ -48,6 +48,15 @@
 namespace dust {
 namespace random {
 
+/// Jump the random number state forward by a number of steps equal to
+/// the square root of the sequence length.  The xoshiro256 generators
+/// have a sequence length of 2^256 and so a jump is equivalent to
```
johnlees, line 53:
```suggestion
/// have a sequence length of 2^256 and so each of these jumps is equivalent to
```

scripts/docs_build
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
johnlees, line 1: Is the use/purpose of these three scripts noted somewhere?

sphinx/conf.py
```diff
@@ -0,0 +1,71 @@
...
+# Add any paths that contain custom static files (such as style sheets) here,
+# relative to this directory. They are copied after the builtin static files,
+# so a file named "default.css" will overwrite the builtin "default.css".
+html_static_path = ['_static']
+
+breathe_projects = { "dust": "doxygen/xml" }
+breathe_default_project = "dust"
+breathe_show_define_initializer = True
+
+cpp_id_attributes = ["__host__", "__device__", "__nv_exec_check_disable__"]
```
johnlees, line 65: I'm sure this was easy to figure out 😛

sphinx/examples/Makefile
```diff
@@ -0,0 +1,14 @@
+SRC = $(wildcard *.cpp)
+OUT = $(patsubst %.cpp,%.out,$(SRC))
+BIN = $(patsubst %.cpp,%,$(SRC))
+
+all: $(BIN) $(OUT)
+
+%: %.cpp
+	$(CXX) -O0 -I../../inst/include -o $@ $<
```
johnlees, line 8: why `O0`?

sphinx/random-distributions.rst
```diff
@@ -0,0 +1,47 @@
+Distributions
+=============
+
+Our distribution support will grow over time. If you need a distribution that is not yet supported please `post an issue <https://github.com/mrc-ide/dust/issues/>`_.
```
johnlees, line 4:
```suggestion
Our distribution support will grow over time. If you need a distribution that is not yet supported please `post an issue <https://github.com/mrc-ide/dust/issues/>`__.
```
I believe is best, otherwise it makes an internal link or something

johnlees (changes requested):

sphinx/examples/Makefile
richfitz: fastest compilation :)

README.md
```diff
@@ -35,6 +35,8 @@ And several on the random number generator, around which dust is built:
 * `vignette("rng_algorithms")` on the details of algorithms used to sample from some distributions
 * `vignette("rng_package")` on using the generators from other packages
 
+The C++ API is documented in a separate [set of documentation](cpp)
```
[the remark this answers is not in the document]
richfitz: yeah, that's not going to be possible with this setup. As I was writing a blog post though I thought of a better way, and will implement that after. It will remove sphinx and breathe entirely from this system and create a set of additional markdown files that should easily be incorporated into the usual doc structure

developing.md
```diff
@@ -29,3 +33,11 @@ As the project has become more complex, keeping the headers under control has be
 * Only include cpp11 (or R) files from files within `dust/r/`; within these files throw errors only with `throw`, not with `cpp11::stop` (these errors will be correctly caught).
 
 The script `scripts/check_headers` will validate that headers are self contained and that only interface headers include `cpp11` files (directly or indirectly).
+
+## Documentation
+
+There are lots of places to consider putting documentation
+
+* Following R package standards, all user facing functions must be documented and have examples, we use roxygen for this (generates files in `man/`). However, because most of the work in dust
```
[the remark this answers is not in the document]
richfitz: no idea what that was going to say

scripts/docs_build
richfitz: not really, and I hope they're temporary. all these scripts are a bit mysterious I expect :)

sphinx/conf.py
richfitz: not too bad - it fails until you list them all here

sphinx/index.rst
```diff
@@ -0,0 +1,10 @@
+dust C++ documentation
+======================
+
+.. toctree::
+   :maxdepth: 2
+   :caption: Contents:
+
+   random
+
+This set of pages describes the ``dust`` C++ API in more detail than that discussed in the main ``dust`` documentation.  Principally it covers the random number library.
```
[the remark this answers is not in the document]
richfitz: I have no idea how this whole set up works evidently and am not sure where that would go? In light of the other comments though it might not be worth getting too attached to the rst/sphinx docs

johnlees (approved):

PR #359, opened 2022-01-12
Allow underscores in dust::name and similar

### LINKED ISSUE #358: Allow dust to compile models with underscore names
Given we randomly generate a package name this should be dooable - see https://github.com/mrc-ide/odin.dust/issues/98 for context

richfitz:
See https://github.com/mrc-ide/odin.dust/issues/98 for the trigger.  We just fall back to using `dust` for the dll name which should work fine.  For packages, this is not a problem at all.

Fixes #358 
Fixes https://github.com/mrc-ide/odin.dust/issues/98

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: dust
 Title: Iterate Multiple Realisations of Stochastic Models
-Version: 0.11.21
+Version: 0.11.22
 Authors@R: c(person("Rich", "FitzJohn", role = c("aut", "cre"),
                     email = "rich.fitzjohn@gmail.com"),
              person("John", "Lees", role = "aut"),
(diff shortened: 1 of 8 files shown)


chvandorp:
Hi Rich, I installed the `i358-underscore` branch and tried compiling a model with `odin.dust::odin_dust`. Underscores in filenames now work fine. However, I was able to come put with some other annoying filenames, but perhaps that's for another issue.
```R
odin.dust::odin_dust("odin_dust_test(1).txt") ## causes error in R script
odin.dust::odin_dust("odin_dust.test.txt") ## causes C++ compilation error
odin.dust::odin_dust("odin_dust_test_missing.txt") ## if the file is missing, the warning does not show the filename
```
Thanks!

richfitz:
Thanks for these examples @chvandorp; I've fixed these in odin (see https://github.com/mrc-ide/odin/pull/254, `allow-parens`, version 1.3.3). With that installed, all 3 of your examples behave much better

weshinsley (approved):

PR #360, opened 2022-01-21
Allow partial particle filter running/running from deterministic models

### LINKED ISSUE #177: Filter should allow partial running

richfitz:
Two related (though actually different) changes here to allow:

* running the particle filter over a range of data; from the current step to some time within the data set
* running the deterministic "filter"; this allows the compare function to be parallelised when running multiple regions, and will hopefully reduce some overhead in sircovid's use

Fixes #177


inst/include/dust/filter.hpp
```diff
@@ -37,9 +58,18 @@ filter(T * obj,
...
-    obj->compare_data(weights, d.second);
+  auto d = obj->data().cbegin();
+  const auto d_end = obj->data().cend();
+
+  while (d->first <= obj->step() && d != d_end) {
+    d++;
+    state.trajectories.advance();
+  }
+
+  for (; d != d_end; ++d) {
```
johnlees, line 69: I've never seen a loop like this before, not to say it's wrong.
Would it not be more typical to have
```
for (auto it = d; it != d_end; ++it)
```

inst/include/dust/filter.hpp
```diff
@@ -27,7 +44,11 @@ filter(T * obj,
 
   if (save_trajectories) {
     state.trajectories.resize(obj->n_state(), n_particles, n_data);
-    obj->state(state.trajectories.value_iterator());
+
+    if (obj->step() <= obj->data().begin()->first) {
```
johnlees, line 48: Comment to explain what this is checking? The second term is a bit complex

inst/include/dust/filter.hpp
```diff
@@ -69,16 +100,86 @@ filter(T * obj,
...
+    }
   }
 
   return log_likelihood;
 }
 
 
+template <typename T>
+std::vector<typename T::real_type>
+deterministic(T * obj,
```
johnlees, line 119: Function looks very similar to `stochastic`. What were your thoughts vs a single function with a `deterministic` boolean and some if blocks?

johnlees: If keeping separate, then I think if any changes are made above (comment and loop init) they should also be done here

inst/include/dust/filter_state.hpp
```diff
@@ -133,18 +133,25 @@ class filter_trajectories_host {
...
+      auto it_ret = ret + i * n_state_particles;
       for (size_t j = 0; j < n_particles_; ++j) {
         const size_t idx = *(it_order + index_particle[j]);
         index_particle[j] = idx;
         std::copy_n(it_value + idx * n_state_, n_state_,
                     it_ret + j * n_state_);
       }
     }
+
+    if (offset_ < n_data_) {
```
johnlees, line 150: Comment on why this is needed/when it is used?

inst/include/dust/gpu/filter.hpp
```diff
@@ -38,17 +43,26 @@ filter(T * obj,
...
-  for (auto & d : obj->data()) {
+  auto d = obj->data().cbegin();
+  const auto d_end = obj->data().cend();
+
+  while (d->first <= obj->step() && d != d_end) {
+    d++;
+    state.trajectories.advance();
+  }
+
+  for (; d != d_end; ++d) {
```
johnlees, line 54: as above

inst/include/dust/gpu/filter.hpp
```diff
@@ -28,7 +29,11 @@ filter(T * obj,
 
   if (save_trajectories) {
     state.trajectories.resize(obj->n_state(), n_particles, n_data);
-    state.trajectories.store_values(obj->device_state_selected());
+
+    if (obj->step() <= obj->data().begin()->first) {
```
johnlees, line 33: as above

inst/include/dust/r/dust.hpp
```diff
@@ -446,48 +446,61 @@ cpp11::sexp save_snapshots(const filter_state& snapshots, const T *obj,
...
+                        cpp11::sexp r_step_snapshot,
+                        cpp11::sexp min_log_likelihood) {
   T *obj = cpp11::as_cpp<cpp11::external_pointer<T>>(ptr).get();
   obj->check_errors();
 
   if (obj->n_data() == 0) {
     cpp11::stop("Data has not been set for this object");
   }
 
+  size_t step = std::numeric_limits<size_t>::max();
```
johnlees, line 483: Can't this just be left uninitialised as this value will never be read?

johnlees: Ah wait, I see how it's used. I wonder if `max_step` might be clearer here

tests/testthat/test-filter.R
```diff
@@ -184,3 +184,149 @@ test_that("Validate step_snapshot", {
...
+  for (i in seq_along(ans)) {
+    ans[[i]] <- mod$filter(dat$dat$step[[i]], save_trajectories = TRUE)
+  }
+
+  expect_equal(vapply(ans, "[[", 1, "log_likelihood"), ll)
+
+  ## I don't have a good intuition about why we're correct for the
+  ## state at the end, for each time point, but not for the points in
+  ## the middle; it's very likely that because these are the fully
+  ## sorted trajectories that I just don't understand something.
```
johnlees, line 238: Could easily be an RNG difference?

johnlees: Especially as the deterministic test doesn't have this issue

johnlees (changes requested):

inst/include/dust/filter.hpp
richfitz: this is fairly conventional, c++ is silly

tests/testthat/test-filter.R
richfitz: Ah that's a stale comment, I figured that out

inst/include/dust/filter.hpp
richfitz: these ones could actually be merged you're right (unlike the comment I left in mcstate

inst/include/dust/r/dust.hpp
richfitz: updated that now, definitely clearer

inst/include/dust/filter_state.hpp
richfitz: added a comment on that

inst/include/dust/filter.hpp
richfitz: both a comment and a bit of self-documenting code to tidy it up. it's not bad, but not obvious when within the condition

johnlees (approved):

PR #362, opened 2022-01-26
Add min log likelihood support

### LINKED ISSUE #361: Support for min_log_likelihood
As in mcstate, see https://github.com/mrc-ide/dust/pull/360

richfitz:
Minor tweak to support `min_log_likelihood` within the cpu models.  It could be added to the GPU models too, but that requires fetching the log_likelihood from the device at every step, and so will likely remove too much of the benefit here.

This trick was discovered in the mcstate code (see [here](https://mrc-ide.github.io/mcstate/reference/particle_filter.html)) were we can compute the minimum log likelihood that we would accept in a pmcmc (https://github.com/mrc-ide/mcstate/blob/7485a6a8d37cee73523ef835366ef20ad7123329/R/pmcmc_state.R#L76-L90) under certain conditions (when the likelihood function is the sum of normalised discrete probabilities so that it always decreases).  Doing this can get about a 30% speedup in a pmcmc run because many points can be stopped part way through their runs.

Fixes #361


inst/include/dust/filter_tools.hpp
```diff
@@ -59,16 +59,55 @@ real_type scale_log_weights(typename std::vector<real_type>::iterator w, size_t
...
   }
+
   return std::log(tot / n) + max_w;
 }
 
 
+template <typename real_type>
+bool early_exit(const std::vector<real_type>& log_likelihood,
+                const std::vector<real_type>& min) {
+  for (auto x : log_likelihood) {
```
johnlees, line 81: Should this have a `&` to not copy the vector? I can never remember for these loops

inst/include/dust/filter_tools.hpp
```diff
@@ -59,16 +59,55 @@ real_type scale_log_weights(typename std::vector<real_type>::iterator w, size_t
...
     *wi = std::exp(*wi - max_w);
     tot += *wi;
   }
+
   return std::log(tot / n) + max_w;
 }
 
 
+template <typename real_type>
+bool early_exit(const std::vector<real_type>& log_likelihood,
```
johnlees, line 79: I was a bit confused as to why these are both vectors, but I guess that is for `n_pars > 1`? Could we add a comment on that here maybe

inst/include/dust/r/helpers.hpp
```diff
@@ -350,6 +350,35 @@ std::vector<size_t> check_step_snapshot(cpp11::sexp r_step_snapshot,
...
+      if (n_pars <= 1) { // avoid unfriendly error message
+        cpp11::stop("'min_log_likelihood' must have length 1 (but given %d)",
+                    n_given);
+      } else {
+        cpp11::stop("'min_log_likelihood' must have length 1 or %d (but given %d)",
+                    n_pars, n_given);
+      }
+    }
+    min_log_likelihood.reserve(n_given);
+    for (auto x : r_min_log_likelihood_vec) {
```
johnlees, line 374: Again, should this have a `&`?

tests/testthat/test-filter.R
```diff
@@ -326,3 +315,115 @@ test_that("can partially run filter in deterministic mode", {
...
+  ## first parameter set (arbitrary) and discard first point to align
+  ## with ll calculation:
+  expect_equal(which(res$trajectories[1, , 1, -1] != 0),
+               which(rowSums(ll) >= -300))
+  expect_equal(res$log_likelihood, rep(-Inf, 2))
+  expect_equal(mod$step(), 600)
+
+  ## If we provide two numbers it's the one reached second:
+  mod$update_state(pars = pars, step = 0)
+  min <- c(mean(ll[35:36, 1]), mean(ll[55:56, 2]))
```
johnlees, line 376: Where do these indices come from?

johnlees (approved):

inst/include/dust/filter_tools.hpp
richfitz: Here `x` is an atomic type (`real_type`, i.e., `double`/`float`) so it makes no difference and I think is more typical not to add a reference

tests/testthat/test-filter.R
richfitz: Nowhere, there are ~100 timesteps in that example and I wanted something that failed in the middle at different times for the different parameters

inst/include/dust/filter_tools.hpp
richfitz: yeah, this could do with a comment, will add one

PR #365, opened 2022-04-19
Allow sharing data across parameters

richfitz:
This PR allows a single data set to be used for all parameter sets, which is something we will want for things like speculative mcmc. Previously we assumed that every parameter set had a distinct data set.


inst/include/dust/gpu/kernels.hpp
```diff
@@ -135,6 +136,7 @@ void run_particles(size_t step_start,
                            shared_real,
                            nullptr,
                            false,
+                           false,
```
johnlees, line 139: Helpful to set these bools to named variables to be passed, as it's getting harder to interpret what they mean?

inst/include/dust/r/dust.hpp
```diff
@@ -375,31 +375,31 @@ void dust_set_n_threads(SEXP ptr, int n_threads) {
 }
 
 template <typename T, typename std::enable_if<!std::is_same<dust::no_data, typename T::data_type>::value, int>::type = 0>
-void dust_set_data(SEXP ptr, cpp11::list r_data) {
+void dust_set_data(SEXP ptr, cpp11::list r_data, bool shared) {
```
johnlees, line 378: I think I would prefer `shared_data` throughout, as I am thinking of plain `shared` as shared_pars or the shared L1 cache

inst/template/dust.R.template
```diff
@@ -452,8 +452,12 @@
     ##' parameter sets.
     ##'
     ##' @param data A list of data to set.
-    set_data = function(data) {
-      private$methods_$set_data(private$ptr_, data)
+    ##'
+    ##' @param shared Logical, indicating if the data should be shared
```
johnlees, line 456: `shared_data` here too?

johnlees (approved):

inst/template/dust.R.template
richfitz: I've updated the C++ facing ones, but not this one as the user here should not be thinking about shared memory

inst/include/dust/gpu/filter_state.hpp
```diff
@@ -44,7 +44,7 @@ class filter_trajectories_device : public filter_trajectories_host<real_type> {
                                this->history_value.size() * sizeof(real_type),
                                cudaHostRegisterDefault));
     CUDA_CALL(cudaHostRegister(this->history_order.data(),
-                               this->history_order.size() * sizeof(real_type),
+                               this->history_order.size() * sizeof(size_t),
```
richfitz, line 47: This is an old bug, causes crash when saving trajectories from models that use `float` as `real_type`

inst/include/dust/gpu/launch_control.hpp
```diff
@@ -101,8 +101,8 @@ inline launch_control launch_control_model(size_t n_particles,
 
   const size_t n_pars_effective = n_particles / n_particles_each;
   const int warp_size = dust::gpu::warp_size;
-  const size_t warp_block_size =
-    warp_size * (n_particles_each + warp_size - 1) / warp_size;
+  const size_t warp_block_count = (n_particles_each + warp_size - 1) / warp_size;
+  const size_t warp_block_size = warp_size * warp_block_count;
```
richfitz, line 105: there was an order of operations issue here; for integers `a * b / c` is not always the same as `a * (b / c)` if `b` is not a multiple of `c`

hillalex (approved):

PR #370, opened 2022-08-04
Implement hypergeometric distribution

richfitz:
This was a bit of a rabbit hole, but implements the same algorithm that R's `rhyper` uses, which seems to still be the state of the art. The reference is this article https://www.tandfonline.com/doi/abs/10.1080/00949658508810839 and the rust implementation was very useful in getting it in a modern style https://docs.rs/rand_distr/0.4.1/rand_distr/struct.Hypergeometric.html


tests/testthat/test-rng.R
```diff
@@ -988,3 +988,179 @@ test_that("We can compile the standalone program", {
...
+## implementations, the algorithm in the paper
+## (https://www.tandfonline.com/doi/abs/10.1080/00949658508810839) has
+## been followed pretty exactly. So if we were willing to assume that
+## the R version is correct we can just check that the stream of
+## numbers is exactly the same.
+##
+## We do this twice - once between R's rhyper and a direct
+## reimplementation of the Rust code in R (within helper-rng) using
+## R's RNG stream, then again between our R reimplementation and
+## dust's idiomatic C++ version.
```
hillalex, line 1004: why? why not just test the C++ version directly against rhyper?

hillalex (commented):
I find the actual code here completely unreviewable, but the tests look like they're correct!

tests/testthat/test-rng.R
richfitz: because they use different random number sources - we can't get rhyper to use dust's stream, and can't get dust to use R's stream. I wrote the reference version to accept any source of uniform random numbers to act as a bridge there

src/dust_rng.cpp
```diff
@@ -389,6 +389,50 @@ cpp11::sexp dust_rng_multinomial(SEXP ptr, int n,
...
+  auto n1_vary = check_input_type(r_n1, n, n_streams, "n1");
+  auto n2_vary = check_input_type(r_n2, n, n_streams, "n1");
+  auto k_vary = check_input_type(r_k, n, n_streams, "k");
+
+  dust::utils::openmp_errors errors(n_streams);
+
+#ifdef _OPENMP
+  #pragma omp parallel for schedule(static) num_threads(n_threads)
+#endif
+  for (int i = 0; i < n_streams; ++i) {
```
hillalex, line 413: no tests for the case where the `rng` is distributed/`n_streams > 1`?

richfitz: there aren't for most of the distributions tbh, they probably should be added (though this class is really mostly intended for debugging). I've added one in 06c9ec6 though as you're right there _is_ logic here to be tested

hillalex (approved):

PR #372, opened 2022-09-28
mrc-3537 add negative binomial distribution

hillalex:
Adds the negative binomial distrbution. Depends on gamma distribution implemented here: https://github.com/mrc-ide/dust/pull/373 and the already implemented Poisson distribution. 

Tests are a bit basic here but the code is very simple - as long as gamma and Poisson are correct then this will be too.


tests/testthat/test-rng.R
```diff
@@ -1288,3 +1288,28 @@ test_that("gamma random numbers prevent bad inputs", {
...
+  expect_error(dust_rng$new(1)$nbinomial(1, 10, 0),
+               "Invalid call to nbinomial with size = 10, prob = 0")
+  expect_error(dust_rng$new(1)$nbinomial(1, 0, 0.5),
+               "Invalid call to nbinomial with size = 0, prob = 0.5")
+  expect_error(dust_rng$new(1)$nbinomial(1, 10, 1.5),
+               "Invalid call to nbinomial with size = 10, prob = 1.5")
+  expect_error(dust_rng$new(1)$nbinomial(1, 10, Inf),
+               "Invalid call to nbinomial with size = 10, prob = inf")
+
+  # this is coming out as -2147483648; in general, rng$poisson for sufficiently
```
hillalex, line 1312: some kind of bug in Poisson here?

inst/include/dust/random/nbinomial.hpp
```diff
@@ -0,0 +1,38 @@
...
+    dust::utils::fatal_error(buffer);
+  }
+}
+
+}
+
+template <typename real_type, typename rng_state_type>
+real_type nbinomial(rng_state_type& rng_state, real_type size, real_type prob)
+{
+    nbinomial_validate(size, prob);
```
richfitz, line 31: Can you add an assert that this does not work for CUDA, and a deterministic route (as in the gamma case). If you could move that closing brace too please - probably we should set up clang-format or something at some point

richfitz (changes requested):

inst/include/dust/random/nbinomial.hpp
```diff
@@ -0,0 +1,42 @@
...
+#endif
+    nbinomial_validate(size, prob);
+
+    if (rng_state.deterministic) {
+      return (1 - prob) * size / prob;
+    }
+    return (prob == 1) ? 0 : poisson(rng_state, gamma(rng_state, size, (1 - prob) / prob));
+}
+
+}
```
richfitz, line 39: sorry, I should have noticed this - somehow the `nbinomial` here is within the anonymous namespace, but also being used. I must be wrong about how this is affecting visibility, though this suggests I am right: https://www.learncpp.com/cpp-tutorial/unnamed-and-inline-namespaces/

So how are we accessing dust::random::nbinomial from dust_rng.cpp???

hillalex: Huh, yeah I see what you mean. And looks like I've nested the gamma function in the same way. In `dust_rng.cpp` they are just referenced like `dust::random::nbinomial` etc

richfitz (approved):

PR #373, opened 2022-09-29
mrc-3536 add gamma distribution

hillalex:
Adds draws from the gamma distrbution based on the implementation of `rgamma` in the R source code (https://github.com/wch/r-source/blob/trunk/src/nmath/rgamma.c). The rust implementation used a different algorithm, but using the same one as `rgamma` makes for the nice testing pattern that you came up with for the hypergeometric case.


inst/include/dust/random/gamma.hpp
```diff
@@ -0,0 +1,225 @@
...
+      if (exponential_rand<real_type>(rng_state) >= x)
+        return x;
+    }
+  }
+}
+
+// Generate gamma random number via GD, p 53 of reference.
+template <typename real_type, typename rng_state_type>
+real_type gamma_gd(rng_state_type& rng_state, real_type a) {
+  const real_type q1 = 0.04166669;
```
hillalex, line 54: Codecov is complaining about the method complexity here. It could certainly be refactored but at the moment is very close to the R source implementation (minus a couple of superfluous lines) which is perhaps an advantage in other ways (can easily be verified as equivalent to that implementation).

richfitz (commented):
If you don't mind, I'd rather we reflect the Rust version and https://dl.acm.org/doi/10.1145/358407.358414 as it's more recent than the R version and likely easier to understand and implement, along with the licencing issues. The implementation will hopefully be simple enough that we can check it the same way as for the other distributions (draw a ton of numbers and check some moments - because this is a continuous distribution we can also use ks.test - there are some examples in the tests already

inst/include/dust/random/gamma.hpp
```diff
@@ -0,0 +1,115 @@
...
+             a, b);
+    dust::utils::fatal_error(buffer);
+  }
+}
+
+template <typename real_type, typename rng_state_type>
+real_type gamma_large(rng_state_type& rng_state, real_type a) {
+  real_type d = a - 1.0 / 3.0;
+  real_type c = 1.0 / sqrt(9.0 * d);
+  while(true) {
```
richfitz, line 38:
```suggestion
  while (true) {
```

inst/include/dust/random/gamma.hpp
```diff
@@ -0,0 +1,115 @@
...
+    real_type x = normal<real_type>(rng_state, 0, 1);
+    real_type v_cbrt = 1.0 + c * x;
+    if (v_cbrt <= 0.0) {
+      continue;
+    }
+    real_type v = v_cbrt * v_cbrt * v_cbrt;
+    real_type u = uniform<real_type>(rng_state, 0, 1);
+    real_type x_sqr = x * x;
+    if (u < 1.0 - 0.0331 * x_sqr * x_sqr ||
+      log(u) < 0.5 * x_sqr + d * (1.0 - v + log(v))) {
```
richfitz, line 48:
```suggestion
      std::log(u) < 0.5 * x_sqr + d * (1.0 - v + std::log(v))) {
```

inst/include/dust/random/gamma.hpp
```diff
@@ -0,0 +1,115 @@
...
+      return d * v;
+    }
+  }
+}
+
+template <typename real_type, typename rng_state_type>
+real_type gamma_small(rng_state_type& rng_state, real_type a) {
+  real_type inv_shape = 1 / a;
+  real_type u = uniform<real_type>(rng_state, 0, 1);
+  return gamma_large(rng_state, a + 1.0) * pow(u, inv_shape);
```
richfitz, line 58:
```suggestion
  return gamma_large(rng_state, a + 1.0) * std::pow(u, inv_shape);
```

inst/include/dust/random/gamma.hpp
```diff
@@ -0,0 +1,115 @@
...
+      log(u) < 0.5 * x_sqr + d * (1.0 - v + log(v))) {
+      return d * v;
+    }
+  }
+}
+
+template <typename real_type, typename rng_state_type>
+real_type gamma_small(rng_state_type& rng_state, real_type a) {
+  real_type inv_shape = 1 / a;
+  real_type u = uniform<real_type>(rng_state, 0, 1);
```
richfitz, line 57: for `U(0, 1)` draws use `dust::random::random_real<double>(rng_state)`

richfitz: (There's some degree of C++ API docs here: https://mrc-ide.github.io/dust/cpp/)

inst/include/dust/random/gamma.hpp
```diff
@@ -0,0 +1,115 @@
...
+///
+/// @param rng_state Reference to the random number state, will be
+/// modified as a side-effect
+///
+/// @param a Shape
+///
+/// @param b Scale
+template <typename real_type, typename rng_state_type>
+__host__ __device__
+real_type gamma(rng_state_type& rng_state, real_type a, real_type b) {
```
richfitz, line 82: Rather than gamma(a, b) should we do this as gamma(shape, scale), especially as there are two parameterisations on the wikipedia page!

tests/testthat/helper-rng.R
```diff
@@ -30,25 +31,25 @@ hypergeometric_r <- function(random_real) {
...
 
     x_l <- m - d + 0.5
     x_r <- m + d + 0.5
 
     k_l <- exp(a -
-               lfactorial(x_l) -
-               lfactorial(n1 - x_l) -
-               lfactorial(k - x_l) -
-               lfactorial((n2 - k)  + x_l))
+                 lfactorial(x_l) -
```
richfitz, line 44: We'll get churn here if one of us doesn't reconfigure our indent rules (looks like throughout). No idea which of the two versions is really prevailing style. Not wildly fussed though

tests/testthat/test-rng.R
```diff
@@ -1178,3 +1178,75 @@ test_that("numbers on different streams behave as expected", {
...
+
+  gamma <- rng1$gamma(n, 1, b)
+  exp <- rng2$exponential(n, 1 / b)
+
+  expect_equal(gamma, exp)
+})
+
+
+test_that("can draw gamma random numbers", {
+  ## a > 1
```
richfitz, line 1198:
```suggestion
  ## when a > 1
```

avoid lint false positive (and below)

tests/testthat/test-rng.R
```diff
@@ -1178,3 +1178,75 @@ test_that("numbers on different streams behave as expected", {
...
+  state_f <- rng_f$state()
+  state_d <- rng_d$state()
+
+  expect_equal(rng_f$gamma(n_reps, a, b), a * b,
+               tolerance = 1e-6)
+  expect_equal(rng_d$gamma(n_reps, a, b), a * b)
+
+  expect_equal(rng_f$state(), state_f)
+  expect_equal(rng_d$state(), state_d)
+})
```
richfitz, line 1238: add a test that with a == 1 you get _exactly_ the same stream as you'd expect from exponential?

richfitz (changes requested):
Looking good - coverage change looks like it's unrelated to this PR?

Can you add the distribution to the docs here: https://mrc-ide.github.io/dust/articles/rng.html#supported-distributions

I think moving to shape/scale parameters rather than a/b would make it easier to read. Probably also worth adding a non-deterministic single precision (float) test too as these have hidden bugs before

tests/testthat/test-rng.R
hillalex: this one does that: https://github.com/mrc-ide/dust/pull/373/files#diff-957923b771229f7f58143ed5a4d8b30a42219ad7ec403e0bfc7221ee7c298fe6R1183

src/dust_rng.cpp
```diff
@@ -433,6 +433,46 @@ cpp11::sexp dust_rng_hypergeometric(SEXP ptr, int n,
   return sexp_matrix(ret, n, n_streams);
 }
 
+
+template <typename real_type, typename T>
+cpp11::sexp dust_rng_gamma(SEXP ptr, int n,
+                           cpp11::doubles r_a, cpp11::doubles r_b, int n_threads) {
```
richfitz, line 439: for consistency, probably also want this to be shape, scale, though it will make it even nastier. Not sure how we can generally keep this from getting so ugly while still keeping it flexible (it's mostly just for testing)

tests/testthat/test-rng.R
richfitz: sorry, totally missed that

richfitz (changes requested):
Looking good, sorry one last change for consistency

R/rng.R
```diff
@@ -279,6 +282,20 @@ dust_rng <- R6::R6Class(
...
+    ##' @description Generate `n` numbers from a gamma distribution
+    ##'
+    ##' @param n Number of samples to draw (per stream)
+    ##'
+    ##' @param a Shape
+    ##'
+    ##' @param b Scale
+    ##''
+    ##' @param n_threads Number of threads to use; see Details
+    gamma = function(n, a, b, n_threads = 1L) {
```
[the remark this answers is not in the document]
hillalex: ah yeah, thanks

richfitz (approved):

PR #374, opened 2022-11-02
Update workflows

richfitz:
There are some unrelated changes here too that emerged as build failures:

* we need to use `std::isfinite`, not `R_FINITE` in the rng code so that it can be used outside of R
* looks like with the ubuntu upgrade, we've exposed that `<limits>` was not correctly included in filter_tools.hpp

I've moved to a (much) simpler release actions a the original gh ones (in the actions/ namespace) are archived. It seems to work: 
image

(I've deleted this release now)


.github/workflows/make-release.yaml
```diff
@@ -88,83 +72,32 @@ jobs:
         with:
           path: pkg
 
-      - name: Show directory
+      - name: Create archive of random library
```
hillalex, line 75: I read this like 5 times in confusion before realising that it meant build dust's random library, as opposed to  building any library _at_ random :laughing:

hillalex (approved):

PR #377, opened 2022-11-16
Add metadata parse support

richfitz:
Adds support for parsing an additional bit of metadata, we'll need this to be able to process mode models smoothly from within dust

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: dust
 Title: Iterate Multiple Realisations of Stochastic Models
-Version: 0.12.1
+Version: 0.12.2
 Authors@R: c(person("Rich", "FitzJohn", role = c("aut", "cre"),
                     email = "rich.fitzjohn@gmail.com"),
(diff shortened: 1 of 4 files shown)


hillalex:
So previously we were relying on the `mode` namespace to identify the type?

richfitz:
Previously, dust always used a discrete time type, and mode always a continuous time. But once we merge the two packages we need a way of hinting what time type to use. The approach used elsewhere in dust is to use these weird annotating comments as we can get the "decor" package to report back on the for us

richfitz:
The other piece of the puzzle here is that within odin.dust at the moment we switch on the model type and use either `dust::dust` or `mode::mode` - with this tweak we're along the way towards `dust` being able to sort this out itself

richfitz:
The other piece of the puzzle here is that within odin.dust at the moment we switch on the model type and use either `dust::dust` or `mode::mode` - with this tweak we're along the way towards `dust` being able to sort this out itself

PR #382, opened 2022-11-28
Tweak to fix Poisson on GPU

richfitz:


richfitz:
Original ticket contents, duplicated here:

Both branches, quite weird, but very easy to trigger:

```
gpu <- dust::dust_cuda_options(fast_math = TRUE, profile = FALSE,
                               quiet = FALSE, debug = FALSE)
gen <- odin.dust::odin_dust({
  initial(x) <- 1
  update(x) <- rpois(5)
  initial(y) <- 1
  update(y) <- rpois(15)
}, real_type = "float", gpu = gpu, verbose = TRUE)

mod_cpu <- gen$new(list(), 0, 16, 1L)
mod_gpu <- gen$new(list(), 0, 16, 1L, gpu_config = 0L)

t <- 1:20
y_cpu <- mod_cpu$simulate(t)
y_gpu <- mod_gpu$simulate(t)

# *Both* the big and small path wrong here, wtf
y_gpu[1, , ]
y_gpu[2, , ]
```

Less agressively, (with real_type = "double" and fast_math off) we get the same answer

We use utils::integer_max in both branches, but it's not obvious that could be issue as it just refers to a C constant.

There are lots of casts that have not been made to single precision, but given this is also an issue in double precision it won't be that either.

The structure is slightly different to the binomial version (which separates things into deterministic/stochastic more cleanly and puts the syncwarp in a nicer place) but it's not clear that could be issue.

The *really* weird thing is that all branches, in lockstep go 0, 1, 0, 1 for these draws on both branches...

To run this easily I need to be on my workstation and reboot it so that I can use the graphics device. Probably writing a small standalone example (based on dust-bench) would be best?

hillalex (approved):

PR #383, opened 2022-12-01
Use single precision math functions on GPU

richfitz:
This PR provides wrapper functions for math support that will use single precision on the GPU when `real_type` is `float`. This is necessary because of [this issue](https://stackoverflow.com/questions/68058821/why-does-stdsin-work-in-the-cuda-kernel) where cuda processes `std::log(x)` into `log(x)` which then uses the double-precision math function even when `x` is a float. To get around this we need to use the [single prevision CUDA math functions](https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH__SINGLE.html).

I've done this in a way that will hopefully tidy up our existing numeric.hpp support, progressively. For now the two things can live side by side.

The nice thing about C++ is that this should be totally without performance cost. With `use_sqrt.cpp` containing:

```
#ifdef USE_DUST

#include "dust/random/math.hpp"

double f(double a, double b) {
  return std::sqrt(a) + std::log(b);
}

#else

#include <cmath>

double f(double a, double b) {
  return std::sqrt(a) + std::log(b);
}

#endif
```

```
g++ -I../inst/include -S -C -o use_sqrt_std.s use_sqrt.cpp
g++ -I../inst/include -DUSE_DUST -S -C -o use_sqrt_dust.s use_sqrt.cpp
git diff --no-index use_sqrt_std.s use_sqrt_dust.s
```

gives

```diff
diff --git a/use_sqrt_std.s b/use_sqrt_dust.s
index 7bae51b6..89d28996 100644
--- a/use_sqrt_std.s
+++ b/use_sqrt_dust.s
@@ -1,9 +1,15 @@
        .file   "use_sqrt.cpp"
        .text
+       .section        .rodata
+       .type   _ZStL19piecewise_construct, @object
+       .size   _ZStL19piecewise_construct, 1
+_ZStL19piecewise_construct:
+       .zero   1
+       .text
        .globl  _Z1fdd
        .type   _Z1fdd, @function
 _Z1fdd:
-.LFB246:
+.LFB1440:
        .cfi_startproc
        endbr64
        pushq   %rbp
@@ -26,7 +32,7 @@ _Z1fdd:
        .cfi_def_cfa 7, 8
        ret
        .cfi_endproc
-.LFE246:
+.LFE1440:
        .size   _Z1fdd, .-_Z1fdd
        .ident  "GCC: (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0"
        .section        .note.GNU-stack,"",@progbits
```

so there's no extra function call to worry about


inst/include/dust/random/poisson.hpp
```diff
@@ -77,26 +80,29 @@ real_type poisson_hormann(rng_state_type& rng_state, real_type lambda) {
...
 
   while (true) {
     real_type u = random_real<real_type>(rng_state);
-    u -= 0.5;
+    u -= static_cast<real_type>(0.5);
     real_type v = random_real<real_type>(rng_state);
 
-    real_type u_shifted = 0.5 - std::fabs(u);
-    real_type k = floor((2 * a / u_shifted + b) * u + lambda + 0.43);
+    real_type u_shifted = static_cast<real_type>(0.5) - std::fabs(u);
```
hillalex, line 104: should `std::fabs` be `dust::math::abs`?

inst/include/dust/random/normal_ziggurat.hpp
```diff
@@ -94,7 +93,7 @@ real_type random_normal_ziggurat(rng_state_type& rng_state) {
     const auto i = ziggurat_layer_draw(rng_state, value, n);
     const auto u0 = 2 * int_to_real<real_type>(value) - 1;
 
-    if (std::abs(u0) < y[i]) {
+    if (std::fabs(u0) < y[i]) {
```
hillalex, line 96: is this right or do you want `dust::math::fabsf`?

hillalex (approved):

PR #384, opened 2022-12-07
Add mode compatibility to dust

richfitz:
Adds some stub methods and additional arguments required for working with ode models.

The size of the diff is misleading - this is mostly generated code (i.e., R/cpp11.R, R/dust.R, R/dust_generator.R, src/*.cpp, msaan/dust_generator.Rd are all generated)


inst/include/dust/r/dust.hpp
```diff
@@ -582,6 +590,18 @@ int dust_n_state(SEXP ptr) {
...
+template <typename T>
+void dust_set_stochastic_schedule(SEXP ptr, SEXP time) {
+  if (time != R_NilValue) {
+    cpp11::stop("'set_stochastic_schedule' not supported in discrete-time models");
+  }
+}
+
+template <typename T>
+void dust_ode_statistics(SEXP ptr) {
+  cpp11::stop("'ode_statistics' not supported in discrete-time models");
```
hillalex, line 602: how will this be made conditional on model type?

richfitz: In the first instance I think the dust and mode models will use entirely separate include files (so the mode ones will be copied over into `inst/include/mode`) and then we'll handle this in the subsequent harmonisation of the two sets of includes.

One way of doing it for real will be to use template metaprogramming the same way that we turn things on and off in dust for gpu support:

```c++
template <typename T, typename std::enable_if<std::is_same<dust::no_data, typename T::data_type>::value, int>::type = 0>
void dust_set_data(SEXP ptr, cpp11::list r_data, bool data_is_shared) {
  disable_method("set_data");
}
```

possibly by adding a `time_type` type within the class

hillalex (approved):

PR #387, opened 2023-02-03
Import mode code

richfitz:
Merge after #386, as this contains those commits. Best to merge https://github.com/mrc-ide/mode/pull/46 too as that's in the imported code, though this PR essentially obsoletes that package!

This PR pulls in the mode headers and tests, and wires up running with mode. I've done as little as possible to modify the imported code here, just gets things passing. Later, we have some harmonisation work to do (see the epic for at least some of the details).

The first commit is the copy (plus some minor rewriting), code to do this is included in the commit message. The range https://github.com/mrc-ide/dust/compare/55f70dd...mrc-3956 then includes the changes made to the copied files, and to dust itself, in order to get things working. There's a little churn in 3862572 moving the mode headers to be self contained and use old-style header guards to match dust.

There is lots of things that should be done on top of this - eliminating the mode namespace, moving all code from inst/include/mode to inst/include/dust etc - but we can do that in stages

The coverage hole is because we don't include any compiled ode examples in the package - I'd rather do that down the line.


inst/include/dust/r/dust.hpp
```diff
@@ -152,6 +152,11 @@ template <typename T, typename std::enable_if<std::is_same<size_t, typename T::t
 void dust_initialise(T *obj, bool reset_step_size) {
 }
 
+template <typename T, typename std::enable_if<std::is_same<double, typename T::time_type>::value, int>::type = 0>
```
hillalex, line 155: this syntax is :open_mouth: I am reading the docs and still struggling to grasp what this means

richfitz: yeah, C++ is terrible, I know. This is an example of [SFINAE](https://en.wikipedia.org/wiki/Substitution_failure_is_not_an_error) which is the metaprogramming technique that C++ uses to get generic functions that get resolved at compile time. Basically we want two forms of this (and other) functions where the correct version gets used depending on some trait-like thing in a type

```
// use this version where T::time_type is size_t  - i.e., discrete time models
template <typename T, typename std::enable_if<std::is_same<size_t, typename T::time_type>::value, int>::type = 0>
void dust_initialise(T *obj, bool reset_step_size) {
}

// use this version where T::time_type is double  - i.e., ode models
template <typename T, typename std::enable_if<std::is_same<double, typename T::time_type>::value, int>::type = 0>
void dust_initialise(T *obj, bool reset_step_size) {
  // It does not matter here that a discrete time model will not have an initialise method, as this version of the function never gets instantiated for a that sort of model
  obj->initialise(reset_step_size);
}
```

The parts here are `enable_if` which is does the test - if it fails then no suitable type can be found for the second template parameter causing that template to be passed over for the next one

hillalex (approved):

PR #389, opened 2023-02-10
Store additional rng state for ode models

richfitz:
As is already done for discrete time models, this extra rng state is used in the particle filter (the rest of that support will come later).

~Merge after #388, contains those commits already or see https://github.com/mrc-ide/dust/compare/mrc-3983...mrc-3503 for diff~


inst/include/dust/ode/dust_ode.hpp
```diff
@@ -32,7 +32,7 @@ class dust_ode {
         n_threads_(n_threads),
         shape_({n_particles}),
         m_(model_type(pars)),
-        rng_(n_particles_, seed, false),
+        rng_(n_particles_ + 1, seed, false), // +1 for filter
```
hillalex, line 35: sorry can you remind me what this is about?

richfitz: Sorry - cryptic messages in both the ticket and the description. In order to run a version of the particle filter entirely in compiled code (which gets a small speed boost) we need one extra copy of random number state, in order to draw the numbers used for reordering the particles. We made the call originally that this should not come from any of the particles themselves, so initialise an additional copy of the RNG state

hillalex (approved):

PR #392, opened 2023-02-22
Support for multiple parameter sets at once

richfitz:
This PR allows ode dust objects to be run with multiple parameter sets at once. This turned out not to be too bad in the end.

This PR is really in two parts; the first set of commits (through to 7c60d12) refactors things to bring the ode and discrete time interfaces a bit closer together, then cd08324 actually adds the support; it's not a lot really!

There's some change to the original discrete time model where I back-ported changes from the ode version, but nothing major and no changes to the interface


inst/include/dust/dust_cpu.hpp
```diff
@@ -395,11 +393,10 @@ class dust_cpu {
 
   std::vector<size_t> index_;
   std::vector<dust::particle<T>> particles_;
-  std::vector<dust::shared_ptr<T>> shared_;
```
hillalex, line 398: I can't work out what work this vector was doing before?

hillalex (approved):

inst/include/dust/dust_cpu.hpp
richfitz: Seemingly nothing! I believe this was never removed from the class from some long-passed refactor?

PR #393, opened 2023-02-22
Support for running ode models in deterministic mode

richfitz:
This is less interesting than for the discrete time models, but it's something the covid folks have used to help tune their models. Basically we wire up the rng engine to only return the expectation of any stochastic draw - here that only affects the stochastic schedule so it's less interesting. However, most of the work is basically done, and this helps complete the interface.

Merge after #392 as this branch contains those commits


inst/include/dust/dust_cpu.hpp
```diff
@@ -338,12 +338,6 @@ class dust_cpu {
     n_threads_ = n_threads;
   }
 
-  // NOTE: it only makes sense to expose long_jump, and not jump,
```
richfitz, line 341: This is unused, leftover from some previous idea I think

hillalex (approved):

PR #397, opened 2023-03-03
Avoid integer overflow in binomial

richfitz:
Note that this is still a potential issue in multinomial (mrc-4053) and hypergeometric (mrc-4051) but these are both slightly more fiddly and less used.

Reported via email by a user in Norway. The test case captures the issue fairly well. I originally tried handling everything in the brts branch (which already does not use integers anywhere) but that does not behave well for low np (there was consistent bias even as the number of draws increased). So I've added a switch on the size of `n` between two forms templated on an integer type, on the principle that things will certainly be faster with i32 in the most common case.


tests/testthat/test-rng.R
```diff
@@ -158,6 +158,26 @@ test_that("Binomial random numbers prevent bad inputs", {
 })
 
 
+test_that("provide nice feedback on integer overflow", {
```
hillalex, line 161: Think this test is wrongly named!

richfitz: aha, thanks. Originally I was going to throw a nice error but then worked out I could just get rid of  the problem. I'll rename that...

hillalex (approved):

PR #403, opened 2023-04-14
Alternative "big lambda" implementation for Poisson random numbers

richfitz:
This avoids the limitation in the Hormann algorithm. I tried just removing the limit but it locks up pretty quick; the paper suggests that's due to rounding error in the approximation used in the tails.

The rust library uses rejection sampling based on a Cauchy distribution, which is fairly trivial so I've rolled it in here too.

Annoyingly this approach does not work well in single precision, with inflated variance in the samples (so the ratio of `var(Posson(lambda))/lambda` grows as lambda grows, even though `mean(Posson(lambda))/lambda` is close to 1. I can't work out what the cause of this is so I've just prevented sampling where it's obviously wrong. I might try and implement another algorithm later, but we don't actually need this on a GPU for anything at the moment


inst/include/dust/random/poisson.hpp
```diff
@@ -136,10 +137,57 @@ real_type poisson_hormann(rng_state_type& rng_state, real_type lambda) {
   return x;
 }
 
+__nv_exec_check_disable__
+template <typename real_type, typename rng_state_type>
+__host__ __device__
+real_type poisson_cauchy(rng_state_type& rng_state, real_type lambda) {
+  // The algorithm as in the dust rand_distr crate
```
weshinsley, line 144:
```suggestion
  // The algorithm as in the rust rand_distr crate
```

inst/include/dust/random/poisson.hpp
```diff
@@ -136,10 +137,57 @@ real_type poisson_hormann(rng_state_type& rng_state, real_type lambda) {
...
+  // https://rust-random.github.io/rand/src/rand_distr/poisson.rs.html
+  // using the fat tails of a Cauchy distribution to do generate
```
weshinsley, line 146:
```suggestion
  // using the fat tails of a Cauchy distribution to generate
```

weshinsley (commented):
Great - think I mostly understand it! 
I found two typos in comments and proposed corrections...!
We could also in another PR just squash the couple of lines in coverage.

weshinsley (approved):

PR #404, opened 2023-04-14
Fix integer overflow in hypergeometric

richfitz:
As reported by @cwhittaker1000, this fixes an overflow issue in the hypergeometric sampling, caused by us using integers. The issue is we take the product of two integers early on, and that can overflow - causing either incorrect results or an infinite loop!

The solution here mirrors the approach in the binomial sample where we take reals everywhere and never actually validate that, rounding just before passing to the stochastic function


hillalex (approved):

weshinsley:
Looks good to me - I might intuitively have gone for a fix involving a bigger int type, rather than converting to real - but I don't have a great feel for the implications either way here.

PR #406, opened 2023-05-24
Another corner case in binomial draws, related to mrc-4246

richfitz:
This PR is a new approach to the fix, deriving the mystery random number in the R sources


weshinsley (approved):
Right - looks good

The change with "large_p" on 216/224/236 look optional, but fine either way - perhaps more clear having the explicit >0.5 written in each...

Issue #399, opened 2023-03-12
Reason for upper limit on poisson parameter?


OthmanElHammouchi:
Hi!

First of all: thank you very much for your work on this library, I'm finding it very useful for my master's thesis. One difficulty I've encountered, is the upper limit you've placed on the lambda parameter in the Poisson distribution. Is there any special reason for this? Is it a limitation of the algorithms you're using to generate the samples? I've written a little work-around for the moment, I could try to contribute it if you're open to this.

richfitz:
I believe this is the limit mentioned in the paper and we need to follow up to make sure what the correct workaround is on our side. The check was introduced here: https://github.com/mrc-ide/dust/pull/372 and I had thought there was some discussion there about it, but it must have been somewhere else.

So all we need is either a source saying the algorithm is valid with larger lambda than that (in which case we can remove or update the check) or a different algorithm to fall back on for very large lambda.

richfitz:
The PR #403 has the reason for the issue and a fix, should be merged next week

PR #408, opened 2023-05-25
Fall back on fp64 for large Poisson in fp32 mode

richfitz:
See comment for the context here. This allows us to always run, and always give correct results, at the cost of potentially degraded GPU performance

Context here: https://github.com/mrc-ide/dust/pull/403, https://github.com/mrc-ide/dust/issues/399 and [mrc-4287](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-4287/very-large-poisson-in-single-precision)


weshinsley (approved):
Excellent

PR #409, opened 2023-06-01
Allow control over compiler options

richfitz:
This PR allows full control over compiler options passed through to dust. It's a bit of a pain because this interacts with a user's Makevars file and the order that compilers accept options matters. But this should do the main job for now, which is to allow Neil to enable "fast math" when building odin.dust models


tests/testthat/test-interface.R
```diff
@@ -434,6 +434,30 @@ test_that("change C++ standard compilation", {
...
+  expect_match(txt, "PKG_CXXFLAGS=.* -O3 -ffast-math", all = FALSE)
+})
+
+
+test_that("pass any other options", {
+  skip_on_cran()
+  filename <- dust_file("examples/walk.cpp")
+  path <- dust_generate(filename, quiet = TRUE, mangle = FALSE,
+                        compiler_options = c("--enable-carrot",
+                                             "--no-turnip"),
```
r-ash, line 453: :carrot: :+1:

r-ash (approved):
This looks good! I don't have much knowledge of this package but looks like those compiler options are making it through nicely

PR #415, opened 2023-08-30
Add basic interpolation support

richfitz:
Just piecewise constant and linear (not spline) and with no special help for working with multivariate ys.

I've just blindly copied this in from the js version we wrote based on previous work (https://github.com/reside-ic/interpolate) for the spline version, but the immediate need is the simpler linear version.

This is then used in https://github.com/mrc-ide/odin.dust/pull/136


weshinsley (approved):
Tested on win, and we talked through the maths part - all looks good to me!

PR #420, opened 2025-03-20
Remove the call to `std::min` from binomial generator.

plietar:
`std::min` is not available when building for CUDA devices, and using it leads to undefined results. We have our own version of it in `dust::math::min` we can use instead, which is already marked as `__host__ __device__`.

This bug had been unnoticed because of the `__nv_exec_check_disable__` annotation that is used throughout the file. I've now removed the annotation to avoid future bugs.

I've also made the error formatting conditional on `!__CUDA_ARCH__`, since `sprintf` isn't available either, which was causing warnings.

There's a good chance other generators need a similar treatment - I'll get to those as I need them.

### DIFF
--- a/inst/include/dust/random/binomial.hpp
+++ b/inst/include/dust/random/binomial.hpp
@@ -33,7 +33,6 @@ real_type fast_pow(real_type x, int_type n) {
   return pow;
 }
 
-__nv_exec_check_disable__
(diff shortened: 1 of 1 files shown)


plietar:
@richfitz the PR is slightly different than my accidental master push, replacing the ternary operator by `dust::math::min`.

PR #421, opened 2025-08-12
Support gamma on CUDA.

plietar:
This worked out of the box simply by adding the right annotations. I've tested this on all three parts of the domain (`shape < 1`, `shape == 1` and `shape > 1`) and get results that are equivalent on the CPU and GPU (given the same starting seed).

Oddly enough I do get very occasional insignificant floating point differences between the two, despite using doubles on both. I suspect this is down to differences in the some of the math functions. By comparison, I get no difference at all on the uniform random number generator.


richfitz (changes requested):
I've seen these minor differences before, that was part of the beginning of the end of hoping we could get identical results on the cpu and gpu based code.  Can you bump the artefact action version on those failing actions - I think they'll run ok after that

richfitz (approved):

